PR #5, opened 2023-04-30
1. Fix the missing dependencies for build and test

mzy2240:
Feel free to change the dependency version if you think it is too tight.


ramedina86:
Very useful, thanks!

PR #7, opened 2023-04-30
[Performance] Bump dependency version and use Plotly.js-dist-min

mzy2240:
The main drive is to use the latest Vite 4 for a better development experience. The ui build time has been greatly reduced, and the bundle size is slightly better.

master
```shell
✓ built in 38.33s
```

this PR
```shell
✓ built in 25.38s
```


mzy2240:
Another profiling: reduced the build time from 13s to 11s in an M1 Max Macbook Pro.

ramedina86:
Great stuff, plotly.js-dist-min is much smaller and Plotly is a key dependency. Using Vite 4 makes sense, thanks for the benchmarks. By the way, I created a dev branch to work on the next release's changes, mainly so that README.md and docs aren't inconsistent.

PR #18, opened 2023-05-05
overwrite template versions with current streamsync version

Wazarr94:
quick pr to fix the outdated versions in the templates provided
not sure if removing the strings in ui.json is needed or not, I let you decide.

### DIFF
--- a/apps/default/ui.json
+++ b/apps/default/ui.json
@@ -1,149 +1,149 @@
 {
-   "metadata": {
-      "streamsync_version": ""
-   },
-   "components": {
-      "root": {
-         "id": "root",
-         "type": "root",
-         "content": {
-            "appName": "My App"
-         },
-         "parentId": null,
-         "position": 0,
-         "handlers": {},
-         "visible": true
-      },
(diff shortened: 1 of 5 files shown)


ramedina86:
I'm sorry, I can't accept this PR, as the mechanism is working as intended. I really appreciate you wanting to contribute to the project, but I suggest discussing things beforehand to the best possible use of your time (unless it's a small, clear problem like the one you detected earlier).

Once components are updated, the current version of the framework is saved in `ui.json` via `AppRunner.update_components`. 

The objective of these mechanism is to, in the future, prevent breaking changes by applying version-specific logic. For example, if the `content` in components is renamed to `properties` in version 0.3, and I load a version made with 0.1.6, the software can identify that it's dealing with an older version, change `content` to `properties`, then trigger a component update and reflect that the `ui.json` is now 0.3-compatible.

The reason why the templates mention older versions is because they were created with older versions and haven't been modified ever since. They're still fully compatible though, as there haven't been any major changes.

Wazarr94:
got it, thats alright this helped me understand a bit more the project

PR #22, opened 2023-05-08
Add Tailwind/UnoCSS support

Wazarr94:
Filling this draft PR to add Tailwind support.
I ended up using UnoCSS, a superset of Tailwind, because Tailwind doesn't support runtime. This lead to hacky things that could end up in having to add tailwind classes versions after version.

You can write Tailwind and it will act normal, but you also have access to the names that UnoCSS recognizes. For instance, "opacity-80" can be rewritten as "op80". It's an example among many others.

Like I said in the issue, colliding names are ok, the styles defined in Vue will be prioritized over the UnoCSS styles.
I let you continue the PR as the rest of the development will be dictated by UI choices. Here's what I achieved with the current state:

image

Useful documentation links:
- [Tailwind docs](https://tailwindcss.com/docs/installation)
- [UnoCSS interactive docs](https://unocss.dev/interactive/)
- [UnoCSS Playground](https://unocss.dev/play/) -> maybe it could be added in the UI so people code their component in the playground and import it in the streamsync editor afterwards


ramedina86:
Hi, as I mentioned before we need to find the way to dynamically import this. **If this isn't possible, I'd rather not have it**, as it bloats not only the builder but also the runtime; non-Tailwind users would be penalised for this. Also, package imports should be managed; they shouldn't be loaded using script.

PR #88, opened 2023-09-01
fix: typo in sessions.md

jzcruiser:
the code for "Users without the header x-success" should be:

if headers.get("x-success") is None


ramedina86:
Thank you for correcting this!

Issue #137, opened 2023-09-22
add support of python debugger


FabienArcellier:
I really enjoy the experience with streamsync. The tech feel awesome. Only think that keep me back was missing the python debugger. I think it can be done easily by doing a simple change.

I have replace the function ``_execute_user_code`` in ``AppProcess`` to validate if it works, by the one below.

```python
def _execute_user_code(self) -> None:
  """
  Executes the user code and captures standard output.
  """

  import streamsync
  from contextlib import redirect_stdout
  import io

  streamsyncuserapp = sys.modules.get("streamsyncuserapp")
  if streamsyncuserapp is None:
      raise ValueError("Couldn't find app module (streamsyncuserapp).")

  with redirect_stdout(io.StringIO()) as f:
      script_path = os.path.join(self.app_path, "main.py")
      code_block = compile(open(script_path).read(), script_path, 'exec')
      exec(code_block, streamsyncuserapp.__dict__)

  captured_stdout = f.getvalue()

  if captured_stdout:
      streamsync.initial_state.add_log_entry(
          "info", "Stdout message during initialisation", captured_stdout)
```

The debugger is triggered

![image](https://github.com/streamsync-cloud/streamsync/assets/159559/92f142d6-19c5-48f9-bce6-29f4898cf764)

Is there any way to contact you (discord ??? mine is fabien7751) to discusson on how I could make this work properly with your design ?

PasVegan:
Hi,
I also have trouble profiling the app code, I think making the debugger work will solve the problem with profiling ?

FabienArcellier:
@PasVegan Could you tell me which tool you are using for profiling ? or do you have a snippet you are using ? I can try to reproduce with the snippet I have changed.

ramedina86:
Hi Fabien, thanks for bringing attention to this and finding a solution.

If I understand correctly, the issue was that executing a string doesn't link the reference back to the file so the debugger can't work. By replacing the execute by a real file and not a string, this allows it to work. Can you please confirm my understanding is correct?

Initially, Streamsync allowed you to run code without executing it, but this has been deprecated and hidden. It was confusing. So right now the change you made makes sense. I'd like to clean up the AppRunner a bit to remove the references to "run_code" and implement this. I'm guessing this will also improve the stack traces which currently read e.g. "<string> line 181".

If you want, feel free to submit a PR against the "dev" branch with this and I'll do a bit of cleanup afterwards.

We do have a very new Discord server, I see you that found it!

[...]

PR #139, opened 2023-09-23
compile python user code before running it to allow interactive debugging

FabienArcellier:
solve #137 

This PR implements support for the python debugger. This allows you to add a breakpoint in the code and explore the running context.

I didn't have to introduce new tests. There was already a test that cover

The file name appears in error messages when the user changes the python code and breaks it.

image


ramedina86:
This is a great enhancement and useful beyond the debugger, thanks for submitting!

PR #155, opened 2023-10-08
add custom component for visualizing geo locations on a map using lea…

yangt:
I have tried to make a custom component for rendering geo locations using Leaflet.
I had some issues with rendering the icons when I render the markers when I put the .css and .js files from custom_components_dist and it still needs to be fixed. But the basemap and the points works well, maybe someone with better vue skills can help

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -11,6 +11,7 @@
 				"@apache-arrow/ts": "^12.0.1",
 				"@monaco-editor/loader": "^1.3.3",
 				"arquero": "^5.2.0",
+				"leaflet": "^1.9.4",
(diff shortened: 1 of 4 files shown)


ramedina86:
Hi yangt, apologies for the delay and thanks for your contribution. Unfortunately the idea for custom components isn't to submit against the core repository, but rather to keep it in your own and make them available to someone that might need them. Hence, I'll close this PR.

However, I'm trying to assemble a list of custom components, which will later turn into a "community page" once it's more developed. If you'd like me to link a repository with the custom component, please let me know and I'd be happy to share it here (pinned Discussions message, Discord) and eventually the community page.

PR #166, opened 2023-10-23
install streamsync extensions with pip

FabienArcellier:
A developer can share streamsync extensions as a python package. Components appear in streamsync after a user installs them from ``pip``.

Packages are scanned when the extension starts with ``streamsync_xxxxxx``. The goal is to recognize components added in streamsync from a ``pip install streamsync_xxxxxxxx``.

package demo on pypi: https://pypi.org/project/streamsync_bubblemessage/
package demo/template on github: https://github.com/FabienArcellier/streamsync_bubblemessage

---

Peek 2023-10-23 09-09

### Design

* the python package has to start with ``streamsync_``
* [the mecanism is compatible with multi app pattern](https://www.streamsync.cloud/custom-server.html#multiple-apps-at-once)
* no conflict resolution if 2 pypi package use the same ``streamsync_`` package. I think pip will show an error during installation

### Discussions

* design a gallery to promote community components
* transfer of template for distribution from my account `FabienArcellier` to `streamsync-cloud`
* the logger usage

### DIFF
--- a/docs/docs/custom-components.md
+++ b/docs/docs/custom-components.md
@@ -173,3 +173,11 @@ Collect the files from `ui/custom_components_dist` and pack them in a folder suc
 ### Try them
 
 The folder containing the generated files, e.g. `my_custom_bubbles`, can now be placed in the `extensions/` folder of any Streamsync project. It'll be automatically detected during server startup.
+
+### Distribute your components
+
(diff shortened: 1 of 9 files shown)


FabienArcellier:
I have to move the `ExtensionManager` into `serve` module. We want to avoid interaction from core module with the filesystem

PR #177, opened 2023-11-20
pagination component for repeater and dataframe

FabienArcellier:
This PR implements a Pagination component. This component allows to split a dataframe or repeater into several pages if it's too large. Pagination is a robust solution for displaying dataframes of several thousands rows.

Peek 2023-11-18 21-45

I've documented its usage on ``hello`` app.

### User.

* user can change page
* user goes to previous page / to next page
* user can change the number of elements per page (optional)
* the user can go to a specific page (optional)

### Developer

* the developer associates the number of records with a server state
* the developer associates the number of records per page with a server state (optional)
* the developer can enable fast movement to a specific page
* the developer can suggest to the user to change the number of pages
* the developer subscribes a page change to a backend event
* the developer subscribes a page size change to a backend event

demo as custom extension : [custom_pagination.zip](https://github.com/streamsync-cloud/streamsync/files/13419349/custom_pagination.zip)

### DIFF
--- a/apps/hello/main.py
+++ b/apps/hello/main.py
@@ -20,7 +20,10 @@ def update(state, session):
     main_df = _get_main_df()
     main_df = main_df[main_df['length_cm'] >= state["filter"]["min_length"]]
     main_df = main_df[main_df['weight_g'] >= state["filter"]["min_weight"]]
-    state["main_df"] = main_df
+    state['main_df'] = main_df
+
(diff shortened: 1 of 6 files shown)


FabienArcellier:
I have to remove the support of url parameter, waiting for methods to manage it implement by @ramedina86. I will create a PR to add it later.

I have to use default value on the parameters instead of init to keep consistent behavior when a user erase some parameters.

PR #197, opened 2024-01-17
Data race in StateProxy.state

aleloi:
when state is changed concurrently. Suggested mitigation: construct a list of the items first, and then iterate. A safer solution would be to make a lock around constructing the list, but that's more work due to the recursive construction in StateProxy. `list(...)` is not atomic in Python and the iterator could change while constructing the list. In practice, this change fixed the problem for me.

  File "/app/venv/lib/python3.11/site-packages/streamsync/app_runner.py", line 191, in _handle_state_enquiry
    mutations = session.session_state.user_state.get_mutations_as_dict()

File "/app/venv/lib/python3.11/site-packages/streamsync/core.py", line 232, in get_mutations_as_dict
   for key, value in self.state.items():
RuntimeError: dictionary changed size during iteration

### DIFF
--- a/apps/hello/main.py
+++ b/apps/hello/main.py
@@ -20,7 +20,10 @@ def update(state, session):
     main_df = _get_main_df()
     main_df = main_df[main_df['length_cm'] >= state["filter"]["min_length"]]
     main_df = main_df[main_df['weight_g'] >= state["filter"]["min_weight"]]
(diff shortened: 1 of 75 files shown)


FabienArcellier:
That's great, could you open the PR to branch `streamsync-cloud:dev` ?

PR #201, opened 2024-01-23
Making existing tests location-agnostic; add conditional check for server static path before mounting

mmikita95:
Updated tests in `test_core.py`, `test_app_runner.py`, and `test_serve.py` to use dynamic paths derived from `test_app_dir` in `__init__.py` of the `tests` directory. This ensures consistent path resolution across different environments.

### DIFF
--- a/src/streamsync/serve.py
+++ b/src/streamsync/serve.py
@@ -376,7 +376,7 @@ def get_asgi_app(
                 1. Run the 'npm run build' script in the 'ui' directory and then restart the app.
                 2. Alternatively, launch a UI instance by running 'npm run dev' in the 'ui' directory.
 
-                Please refer to the CONTRIBUTING.md for detailed instructions.x1b[0m"""
+                Please refer to the CONTRIBUTING.md for detailed instructions.\x1b[0m"""
             )
         )
 
--- a/src/streamsync/serve.py
+++ b/src/streamsync/serve.py
@@ -2,6 +2,7 @@ import asyncio
 import mimetypes
 from contextlib import asynccontextmanager
 import sys
+import textwrap
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 import typing
 from fastapi import FastAPI, Request, HTTPException
@@ -41,7 +42,9 @@ def get_asgi_app(
         app_runner.hook_to_running_event_loop()
         app_runner.load()
 
-        if on_load is not None:
+        if on_load is not None \
+           and hasattr(app.state, 'is_server_static_mounted') \
+           and app.state.is_server_static_mounted:
             on_load()
 
         try:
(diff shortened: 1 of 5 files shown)


mmikita95:
Implemented a check to verify the existence of the server static path before mounting it in the ASGI app. Added an error message guiding users on how to resolve the issue if the path is not found.

src/streamsync/serve.py
```diff
@@ -41,7 +42,9 @@ async def lifespan(app: FastAPI):
         app_runner.hook_to_running_event_loop()
         app_runner.load()
 
-        if on_load is not None:
+        if on_load is not None \
+           and hasattr(app.state, 'is_server_static_mounted') \
+           and app.state.is_server_static_mounted:
```
ramedina86, line 47: Why are we checking `is_server_static_mounted` here? My idea is that loading without `/static` should work similarly to a normal run, but log the error message. Not sure why we're only calling `on_load()` only if this succeeded. Has the lack of static caused any issues somewhere else?

src/streamsync/serve.py
```diff
@@ -359,8 +362,23 @@ async def stream(websocket: WebSocket):
...
     server_static_path = pathlib.Path(server_path) / "static"
-    asgi_app.mount(
-        "/", StaticFiles(directory=str(server_static_path), html=True), name="server_static")
+    if server_static_path.exists():
+        asgi_app.mount(
+            "/", StaticFiles(directory=str(server_static_path), html=True), name="server_static")
+        asgi_app.state.is_server_static_mounted = True
+    else:
+        logging.error(
+            textwrap.dedent(
```
ramedina86, line 371: I didn't know `dedent`, that's really nice. Based on this, please improve code formatting wherever you see it -starting with `serve.py` and the "streamsync v0.2.8" message.

tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+from pathlib import Path
+
+test_app_dir = Path(__file__).resolve().parent / 'testapp'
```
ramedina86, line 3: Smart! That should do it

src/streamsync/serve.py
mmikita95: My idea with this addition was that `on_load()`  message (“_App/Builder is available at 127.0.0.1:5000_”) would be misleading in this particular case, because when server static is not mounted, the app is not accessible at 127.0.0.1:5000 (when opened, it displays a JSON object `{"detail": "Not found"}`) – you’d have either to do `npm run build` to add server static and then restart the app, or to launch the UI separately with `npm run dev` as the contributing doc (and the new error message) suggest – but in this case, the frontend becomes accessible through the localhost link that Vite provides you with, and 127.0.0.1:5000 still displays the same JSON I’ve mentioned previously.

src/streamsync/serve.py
mmikita95: Got it!

PR #204, opened 2024-01-24
Fix: Use escaped keys in StateProxy serialization

mmikita95:
Corrected the use of escaped keys in serializing StateProxy mutations. Replaced instance where unescaped key was used, ensuring consistency and correct handling of keys with dots.


ramedina86:
I'll be merging this now, please submit related tests in a different PR.

Issue #198, opened 2024-01-17
Repater component shows extra components when key names contain dots after mutation handling


aleloi:
Screenshot demonstrating the issue:

![2024-01-17-094355_1341x1133_scrot](https://github.com/streamsync-cloud/streamsync/assets/16827347/46c26714-f679-41b9-8ef7-771108bc95b1)

State mutation handling has a bug. The app runner sends which parts of the state has changed to the UI components. The changes (called mutations) are encoded as a json. Field names in the JSON are interpreted as paths to nested objects. A dot in a field name means a nested object. This has problems if the original StateProxy object had dots in the key names.

I've only encountered that this is a user-facing problem when updating the state for a Repeater object.

This branch contains a minimal reproduction: https://github.com/aleloi/streamsync/tree/issue/repro-of-dots-in-stateproxy-keys-issue. [Link to commit with diff](https://github.com/aleloi/streamsync/commit/5cde692f045c703fbbc0b5785d42b6f0d0059504).

To get the image above, check out the branch and run `streamsync run tests/testapp`. Go to the repeater tab and press 'Re-randomize repeater' button.

The expected result is this:

![2024-01-17-095224_1343x776_scrot](https://github.com/streamsync-cloud/streamsync/assets/16827347/1f52f7da-e4c6-47bf-a1b6-9ff96bc90ebf)

ramedina86:
Thanks for sharing, we'll definitely take a look at this. Internally, the intended behaviour is that dots should be automatically escaped with backslashes i.e. "\." in the generated JSON, when used as part of the name of a state element.

mmikita95:
Managed to pinpoint it, [this](https://github.com/streamsync-cloud/streamsync/pull/204) offers a fix.

mmikita95:
@aleloi fix was merged into dev branch. I personally verified the behavior of Repeater in this particular scenario, and, with the new fix, the issue appears to be resolved. I would greatly appreciate if you could confirm it on your end as well, just to be safe.

aleloi:
Thanks, I built from dev and see that it's fixed for my use case!

PR #206, opened 2024-01-25
Expanding mutation test

mmikita95:
`Adding a class method to count initial mutations for a StateProxy. Checking the amount of mutations per each step of the test to ensure that mutations are properly tracked.`

In regards to our prior discussion with @ramedina86, I found that the most reliable way to avoid issues like #198 in the future is to consistently track how many mutations are produced by the `get_mutations_as_dict` method in `StateProxy`, and compare the amount to the "ground truth" (i.e. actual number of mutations, which is easily deductible). I updated the mutation test to do exactly that. A method that counts expected "initial mutations" was also added, to make the test independent from the state of the dictionary initially passed to `TestStateProxy`.

I will be considering designing other tests as well, and any feedback is appreciated on the way.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -49,6 +49,20 @@ class TestStateProxy:
 
     sp = StateProxy(raw_state_dict)
 
+    @classmethod
+    def count_initial_mutations(cls, d, count=0):
+        """
+        Counts the number of mutations that will be performed for a given dictionary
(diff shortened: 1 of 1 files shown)


ramedina86:
Nice tests! I'm not a hardcore test fanatic, but for these specific matters they're really helpful.

PR #214, opened 2024-01-29
Enabling deletion mutations

mmikita95:
Providing `__delitem__` method for `StateProxy` and `StreamsyncState`, and dict-like `remove()` API referencing the same method. Adding a "-" prefix for state keys meant to be removed at the frontend side when applying for mutated set of `StateProxy`. Enabling frontend to distinguish deletion mutations in `ingestMutations` function of `core/index.ts`.

### DIFF
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -266,7 +266,7 @@ class StateProxy:
                 for child_key, child_mutation in child_mutations.items():
                     nested_key = carry_mutation_flag(escaped_key, child_key)
                     serialised_mutations[nested_key] = child_mutation
-            else:
+            elif f"+{key}" in self.mutated:
                 try:
                     serialised_value = state_serialiser.serialise(value)
                 except BaseException:
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -242,27 +242,20 @@ class StateProxy:
     def escape_key(key):
         return key.replace(".", "\.")
 
-    # TODO This method has side effect of clearing mutations
-    # It should be renamed so it's not confused with a simple getter
-    # extract_mutations
     def get_mutations_as_dict(self) -> Dict[str, Any]:
-        serialised_mutations: Dict[str, Union[Dict,
-                                              List, str, bool, int, float, None]] = {}
+        serialised_mutations: Dict[str, Union[Dict, List, str, bool, int, float, None]] = {}
 
-        # Handle deletions
-        deleted_keys = \
-            {self.escape_key(key)
-             for key in self.mutated
-             if key.startswith("-")}
-        for key in deleted_keys:
-            serialised_mutations[f"{key}"] = None
+        def carry_mutation_flag(base_key, child_key):
+            child_mutation_flag, child_key = child_key[0], child_key[1:]
+            return f"{child_mutation_flag}{base_key}.{child_key}"
 
         for key, value in list(self.state.items()):
             if key.startswith("_"):
                 continue
+            
             escaped_key = self.escape_key(key)
-
             serialised_value = None
+
             if isinstance(value, StateProxy):
                 if value.initial_assignment:
                     serialised_mutations[f"+{escaped_key}"] = serialised_value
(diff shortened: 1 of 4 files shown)


src/streamsync/core.py
```diff
@@ -222,19 +222,40 @@ def __setitem__(self, key, raw_value) -> None:
         self.state[key] = value
         self.apply(key)
 
+    def __delitem__(self, key: str) -> None:
```
ramedina86, line 225: This works for explicitly deleted items. But are we covering the case when someone regenerates a new `StateProxy` like Fabien showed in his code (when he used a dictionary comprehension)? 

For example:

`state["animals"] = { "cat": 1, "dog": 2 }`

and then in the second run:

`state["animals"] = { "dog": 3 }`

Are we somehow detecting this and notifying the front end of the deletion?

I don't think it is, please let me know. Unfortunately I think a more complex and less efficient approach will be needed.

ui/src/core/index.ts
```diff
@@ -126,19 +126,38 @@ export function generateCore() {
 			*/
 
 			const accessor = parseAccessor(key);
+			let lastElementIndex = accessor.length - 1
 			let stateRef = userState.value;
-			for (let i = 0; i < accessor.length - 1; i++) {
+			
+			// Check if the accessor is meant for deletion.
+			const isDeletion = accessor[lastElementIndex].charAt(0) === '-';
```
ramedina86, line 133: Can we instead apply what we discussed earlier i.e. reserving the first character of the accessor to indicate the type of operation? i.e. + addition (perhaps change), - (deletion).

I don't want someone to break it when adding a minus to their key name nor I'd like to implement character escaping logic beyond the dots.

src/streamsync/core.py
mmikita95: Well noticed. I'll verify it and post an update.

mmikita95: I've tested this case, and it seems that currently mutations for nested dictionaries are processed such that they're "nullified" each time and then reconstructed with "new" content. I'm not really sure that this behavior is efficient, but it allows to account for the case you've described. So, yeah, non-explicit removal technically work too. 

Though, an interesting edge case that I've noticed to not work here is when we try to "non-explicitly remove" the whole state at once, i.e. 
```python
def very_strange_handler(state):
    state = {}
```
This simply does not produce any mutations at all, seemingly due to not invoking __setitem__ on neither `StreamsyncState` nor `StateProxy`.

ui/src/core/index.ts
mmikita95: My bad, apologies and thank you for noticing. I actually made it that way, but then got for whatever reason distracted and "removed" it for PR, though I shouldn't have had. I'll fix this.

src/streamsync/core.py
ramedina86: Thanks for looking into this in depth; I appreciate it. It's a complex mechanism and it's easy to miss the edge cases. Let me know whenever you have an update, feel free to take your time. Also if you want to look at it together I'm happy to.

ramedina86:
As you suggested, can you please add some tests to this PR? Mainly to test deletion and the "Fabien code" for key exclusion.

ui/src/core/index.ts
```diff
@@ -126,19 +126,39 @@ export function generateCore() {
 			*/
 
 			const accessor = parseAccessor(key);
+			let lastElementIndex = accessor.length - 1
 			let stateRef = userState.value;
-			for (let i = 0; i < accessor.length - 1; i++) {
+			
+			// Check if the accessor is meant for deletion.
```
ramedina86, line 132: Let's always use an accessor character, not just deletion or addition. Otherwise I think we'd potentially have trouble when someone mutates something called `-myvar`. Plus it looks a bit less hacky at least to me.

Feel free to use the same for addition and mutation (+) or use separate e.g. + for addition, ~ for mutation.

ui/src/core/index.ts
```diff
@@ -126,19 +126,39 @@ export function generateCore() {
...
+			const isDeletion = accessor[lastElementIndex].charAt(0) === '-';
+			const isAddition = accessor[lastElementIndex].charAt(0) === '+';
+			
+			if (isDeletion || isAddition) {
+				// Remove the prefix for processing.
+				accessor[lastElementIndex] = accessor[lastElementIndex].substring(1);
```
ramedina86, line 138: Also, I'm seeing the prefix is in the last element. Do you think it's possible to have it at the start? That way we can first split the key in "op" and "accessor".

For example,

`+myobj.abc.def`

op: `+`
accessor: `["myobj", "abc", "def"]`

mmikita95: It's doable per se, but I think that there're going to be issues with reliably and efficiently "carrying" this identifier from children mutations' `get_mutations_as_dict()` upwards to the first call. I'll see what I can do. Will be keeping you updated on this.

ramedina86: Great point, please look into this and let's weigh in the pros and cons of both and make a decision.

mmikita95: Managed to make it work this way, as per [latest commit](https://github.com/streamsync-cloud/streamsync/pull/214/commits/e2f9fbde939762400c71c3de70d4d9389530f9fb).

ramedina86:
Finally got the opportunity to look into this in depth. Splendid work. You now understand the mechanism much better than I do!

PR #224, opened 2024-02-03
Minor fix for test_mutations

mmikita95:
As discussed, this actually produces two mutations rather than one. I must've overlooked it when submitting, my apologies.


ramedina86:
Thanks for the fix!

PR #229, opened 2024-02-05
Asynchronous handlers support

mmikita95:
I'll have to admit from start that I'm not really well-versed with asynchronous Python, and there's a lot for me to learn. Nevertheless, I've put a lot of considerations and research into this.

This is a working implementation - i.e., defining and using a handler as a `async` coroutine in the app is possible now.

For me though, using `asyncio.run` in this context doesn't seem ideal, as it is not generally advised to use repeatedly. But I wasn't able to make things work with many other configurations, while this does work and survives my manual tests (tests are coming up, by the way!)

I'll be happy to discuss this in more depth and, as always, open for any kind of feedback and suggestions.


ramedina86:
Have you tried `loop.run_in_executor`? Seems to be a natural choice for what we're doing, apparently it's not a strange pattern at all. In fact, the default executor is a new `ThreadPoolExecutor`. Can you please look into that if you haven't and let me know your thoughts?

From the async legend: https://superfastpython.com/asyncio-blocking-tasks/#Example_of_Running_IO-Bound_Task_in_Asyncio_with_run_in_executor

Another resource: https://skipperkongen.dk/2016/09/09/easy-parallel-http-requests-with-python-and-asyncio/

ramedina86:
Also, before I forget, when this is done, can you do me a favour and update the docs?

https://www.streamsync.cloud/event-handlers.html

Please add a new sub heading on async handlers. This is easily done by modifying the markdown. Let me know if you need any help setting up Vitepress for previewing.

mmikita95:
@ramedina86 regarding `run_in_executor` – yeah, I've tried that (in fact, it seemed to me as a very sensible solution indeed), but as I recall I had an issue appearing while returning the future, which has not yet been done – in that specific configuration, it had to be either awaited (a whole other story of asyncing/awaiting every function in the callstack – tried that, broke everything 🥲), or serialized to be sent to frontend (and futures are alas not serializable). With this, there's no such issue. I'm not entirely sure why, and maybe it would be worth it to research this in more depth.

Regarding docs – no problem, sure! I'll see to it very soon.

ramedina86:
I'm not a fan of passing the ThreadPoolExecutor around to the core, as I think it's an AppRunner responsibility. I'd try to return the future from core and maybe intercept it, get the result, before sending the packet back from the server?

src/streamsync/core.py
```diff
@@ -1022,9 +1038,13 @@ def _call_handler_callable(self, event_type, target_component, instance_path, pa
                 arg_values.append(session_info)
 
         result = None
-        with contextlib.redirect_stdout(io.StringIO()) as f:
-            result = callable_handler(*arg_values)
-        captured_stdout = f.getvalue()
+        if is_async_handler:
+            async_handler = self._async_handler_executor(callable_handler, arg_values)
+            thread_pool_future = thread_pool.submit(asyncio.run, async_handler)
```
FabienArcellier, line 1043: You are already in the thread pool executor, why calling the thread_pool again ? Is there a benefits of starting a new thread ?

image

If you don't have to, it will avoid to move it through parameters. I join @ramedina86 on this remark of keeping the beast at the high level.

FabienArcellier: I run your PR with this code on hello app

image

src/streamsync/core.py
```diff
@@ -983,7 +986,18 @@ def _handle_binding(self, event_type, target_component, instance_path, payload)
             return
         self.evaluator.set_state(binding["stateRef"], instance_path, payload)
 
-    def _call_handler_callable(self, event_type, target_component, instance_path, payload) -> Any:
+    async def _async_handler_executor(self, callable_handler, arg_values):
```
FabienArcellier, line 989: I am not an expert on managing async code. I write a proposal that seems to work and allow to keep the code symetry on ``_call_handler_callable``

```python
    def _async_handler_executor(self, callable_handler, arg_values):
        async_callable = self._async_handler_executor_internal(callable_handler, arg_values)
        return asyncio.run(async_callable)

    async def _async_handler_executor_internal(self, callable_handler, arg_values):
        with contextlib.redirect_stdout(io.StringIO()) as f:
            result = await callable_handler(*arg_values)
        captured_stdout = f.getvalue()
        return result, captured_stdout
```

src/streamsync/core.py
```diff
@@ -1022,9 +1038,13 @@ def _call_handler_callable(self, event_type, target_component, instance_path, pa
                 arg_values.append(session_info)
 
         result = None
-        with contextlib.redirect_stdout(io.StringIO()) as f:
-            result = callable_handler(*arg_values)
-        captured_stdout = f.getvalue()
+        if is_async_handler:
+            async_handler = self._async_handler_executor(callable_handler, arg_values)
+            thread_pool_future = thread_pool.submit(asyncio.run, async_handler)
+            result, captured_stdout = thread_pool_future.result()
```
FabienArcellier, line 1044: Following my proposal in EventHandler, here is the final code in ``_call_handler_callable``

```python
        if is_async_handler:
            result, captured_stdout = self._async_handler_executor(callable_handler, arg_values)
        else:
            result, captured_stdout = self._sync_handler_executor(callable_handler, arg_values)
```

src/streamsync/core.py
mmikita95: Originally this was a suggestion by @ramedina86 to use ThreadPoolExecutor here, and I imagine this could give us an advantage in terms of handling multiple calls. Your suggestion does work too indeed. We'd probably have to come up with some sort of benchmark test here, to better understand performance implications behind this. For now, let's indeed stick to this to keep things simpler parameter-passing-wise.

src/streamsync/core.py
mmikita95: This does look cleaner indeed, thank you for your input. I'll implement it as per your suggestion.

FabienArcellier (approved):

Issue #230, opened 2024-02-07
use poetry instead of setup tools as package manager


FabienArcellier:
The migration to poetry is done in an iso-functional way. The goal is to automate the management of the development environment and fix dev dependencies to make working together easier.

[more about why I advocate to use poetry as packaging system](https://dev.to/farcellier/i-migrate-to-poetry-in-2023-am-i-right--115)

### Check

* [x] a user can run `streamsync hello` after installing the package with `pip install streamsync[ds]`
* [x] the user can still install optional dependencies `pip install streamsync[ds]`
* [x] a developper can run `bash ./build.sh`
* [x] files referenced in `MANIFEST.in` are present in the distribution package

### Todo

* [x] configure poetry
* [x] configure poetry to handle virtualenv in `.venv`
* [x] migrate the build.sh
* [x] remove MANIFEST.in because poetry handle all the files from src
* [x] modify CONTRIBUTING.md
* [x] remove the dependency group of `test`, use `build` instead

PR #231, opened 2024-02-07
use poetry instead of setup tools as package manager

FabienArcellier:
### Check
implement #230 

Peek 2024-02-07 08-12

### Check

* [x] a user can run `streamsync hello` after installing the package with `pip install streamsync[ds]`
* [x] the user can still install optional dependencies `pip install streamsync[ds]`
* [x] a developper can run `bash ./build.sh`
* [x] files referenced in `MANIFEST.in` are present in the distribution package

### Todo

* [x] configure poetry
* [x] configure poetry to handle virtualenv in `.venv`
* [x] migrate the build.sh
* [x] remove MANIFEST.in because poetry handle all the files from src
* [x] modify CONTRIBUTING.md
* [x] remove the dependency group of `test`, use `build` instead


src/streamsync/app_runner.py
```diff
@@ -539,7 +539,7 @@ def __init__(self, app_path: str, mode: str):
         self.is_app_process_server_ready = multiprocessing.Event()
         self.is_app_process_server_failed = multiprocessing.Event()
         self.app_process_listener: Optional[AppProcessListener] = None
-        self.observer: Optional[watchdog.observers.Observer] = None
+        self.observer: Optional[PollingObserver] = None
```
ramedina86, line 542: Was this intentional?

src/streamsync/__init__.py
```diff
@@ -1,8 +1,9 @@
+import importlib.metadata
 from typing import Union, Optional, Dict, Any
 from streamsync.core import Readable, FileWrapper, BytesWrapper, Config
 from streamsync.core import initial_state, component_manager, session_manager, session_verifier
 
-VERSION = "0.3.0"
+VERSION = importlib.metadata.version("streamsync")
```
ramedina86, line 6: So the idea is to switch to defining version in the package and then fetching it from the code? Makes more sense I guess

src/streamsync/app_runner.py
FabienArcellier: It is, otherwise `mypy` was complaining in `build.sh`

ramedina86:
Looks good and thanks very much for updating the `CONTRIBUTING.md`, I'll have to consult that too.

@mmikita95 @raaymax Please have a look at this great example of updating docs while making changes in the code... Avoids documentation drift. I expect everyone to help keep docs updated; it's straightforward as it's just modifying the markdown files.

pyproject.toml
```diff
@@ -1,59 +1,53 @@
...
+altair = ">= 5.2.0, < 6"
+httpx = ">=0.26.0, < 1"
 
-[project.scripts]
-streamsync = "streamsync.command_line:main"
+[tool.poetry.extras]
+ds = ["pandas", "pyarrow", "plotly"]
 
-[tool.setuptools.dynamic]
-version = { attr = "streamsync.VERSION" }
```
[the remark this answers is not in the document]
FabienArcellier: In the design choices of poetry, there is a strong choice not to read a python module dynamically like in setuptools. I understand this choice but it leads to a frustrating limit. I prefer the setuptools strategy but impossible to use it properly with poetry.

A [plugin written by the creator of fastapi](https://github.com/tiangolo/poetry-version-plugin) go over this limit but this plugin is incompatible with src layout and the PR to fix it has been open for 3 months and hasn't been merged.

src/streamsync/app_runner.py
ramedina86: Ah nice, must be related to the dependencies I upgraded, thanks for fixing.

src/streamsync/__init__.py
FabienArcellier: That's the recommended way to manage this situation. Another one is to have a utility as bumpversion to update both location. I would avoid this one.

ramedina86: I really like this option.

mmikita95:
Indeed, beautifully done. I'll make sure to commit to that as well.

raaymax (commented):
Nice one :)

PR #232, opened 2024-02-08
Fix for escape sequence

mmikita95:
Escaping existing '\' character with another one, to disable SyntaxWarning.

### DIFF
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -243,7 +243,7 @@ class StateProxy:
 
     @staticmethod
     def escape_key(key):
-        return key.replace(".", "\\.")
+        return key.replace(".", r"\.")
 
     def get_mutations_as_dict(self) -> Dict[str, Any]:
         serialised_mutations: Dict[str, Union[Dict, List, str, bool, int, float, None]] = {}
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -1020,7 +1020,7 @@ class EventHandler:
                 session_info = {
                     "id": self.session.session_id,
                     "cookies": self.session.cookies,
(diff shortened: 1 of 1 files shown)


src/streamsync/core.py
```diff
@@ -243,7 +243,7 @@ def apply(self, key) -> None:
 
     @staticmethod
     def escape_key(key):
-        return key.replace(".", "\.")
+        return key.replace(".", "\\.")
```
FabienArcellier, line 246: In that case, I would advocate to use raw string instead of double quote. It remove all risk of interpret the string.

```python
return key.replace(".", r"\.")
```

https://www.geeksforgeeks.org/python-raw-strings/

FabienArcellier: I would also for this sort of issue encourage to code a unit test that reproduce this issue, then make the fix. I should not require huge effort and it will secure advance behavior from the code.

mmikita95: That's a very good idea — I was mostly using raw strings for regular expressions, and it didn't even come to mind that it could be applicable here. Thank you for noticing and suggesting it. 

Regarding unit tests, I haven't had experience with writing them specifically to catch warnings (as in this case, the behavior of program is expected => "correct"). If this was a case you've had in your practice, I'd gladly pick up an example to learn upon. 

Again, thank you for your input — it surely shows a skill when there's a room found for solid improvement even in such simple things. Glad to be working together! 🙂

ramedina86: What's the use of a test here? I'm not following

FabienArcellier: My bad for the test, I think that was a bad advice. It was to remove a warning not to fix a bug.

PR #238, opened 2024-02-12
docs: added missing images

raaymax:


ramedina86:
Yes! Those were indeed missing and I never got to fix that. Thank you.

@FabienArcellier One thing to keep in mind, when we publish new components we should add screenshots for the docs. The name should match the internal type.

Issue #186, opened 2023-12-20
Mounting app on different path does not seem to work


thondeboer:
Hi, loving the streamsync and somewhat worried that it has been quiet for a while on the release front here...hope you have not lost interest! This is my favorite webapp dev tool since it is sooo easy to get something working...Have also developed a couple of custom components...

anyhoo...

I am trying to mount an app under a different path and followed the directions on the custom server section of the docs, but I am getting a 

```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

In the browser and this in the logs:

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/tdeboer/.cache/pypoetry/virtualenvs/dash-bio-B22Gy4nA-py3.10/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 408, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/tdeboer/.cache/pypoetry/virtualenvs/dash-bio-B22Gy4nA-py3.10/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "/home/tdeboer/.cache/pypoetry/virtualenvs/dash-bio-B22Gy4nA-py3.10/lib/python3.10/site-packages/fastapi/applications.py", line 1106, in __call__
    await super().__call__(scope, receive, send)
  File "/home/tdeboer/.cache/pypoetry/virtualenvs/dash-bio-B22Gy4nA-py3.10/lib/python3.10/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/tdeboer/.cache/pypoetry/virtualenvs/dash-bio-B22Gy4nA-py3.10/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/tdeboer/.cache/pypoetry/virtualenvs/dash-bio-B22Gy4nA-py3.10/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/tdeboer/.cache/pypoetry/virtualenvs/dash-bio-B22Gy4nA-py3.10/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
[...]

PR #241, opened 2024-02-14
fix: fastapi lifespan management with multiple apps at once

FabienArcellier:
When streamsync is run by a FastAPI application as a subapplication, its kernel does not start. FastAPI ignores the lifespan associated with a sub-application. It fix #213 & #186.

The documentation [Multiple apps at once](https://www.streamsync.cloud/custom-server.html#multiple-apps-at-once) is currently broken.

This PR provides a helper `streamsync.serve.lifespan` for the FastAPI application that scans apps mounted in FastAPI and identifies which are streamsync instances. The kernel of each application is started.

Peek 2024-02-14 20-53


ramedina86:
Incredible work with solid tests 💯

PR #242, opened 2024-02-15
fix: introducing a fixture for proper teardown

mmikita95:
Looks quite hellish in the diff, but essentially this adds a fixture to properly initiate an AppRunner (with dynamic parameters), and to shut it down during test teardown, to address the issue with AppRunner not being shut down if the test fails.

### DIFF
--- a/tests/test_app_runner.py
+++ b/tests/test_app_runner.py
@@ -6,8 +6,22 @@ from streamsync.app_runner import AppRunner
 import pytest
 from streamsync.ss_types import EventRequest, InitSessionRequest, InitSessionRequestPayload, StreamsyncEvent
 import asyncio
+import contextlib
 
 from tests import test_app_dir
+
+
+@pytest.fixture
+def setup_app_runner():
+    @contextlib.contextmanager
+    def _manage_launch_args(app_dir, app_command):
+        ar = AppRunner(app_dir, app_command)
+        try:
+            yield ar
+        finally:
+            ar.shut_down()
+    return _manage_launch_args
+
 
 class TestAppRunner:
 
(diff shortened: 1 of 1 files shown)


ramedina86:
Nice. Thanks for seeing something that's wrong and doing something about it.

Yes, looks like the diff is mainly the result of the tabs for the context managers.

tests/test_app_runner.py
```diff
@@ -6,9 +6,23 @@
 import pytest
 from streamsync.ss_types import EventRequest, InitSessionRequest, InitSessionRequestPayload, StreamsyncEvent
 import asyncio
+import contextlib
 
 from tests import test_app_dir
 
+
+@pytest.fixture
+def setup_app_runner():
```
FabienArcellier, line 15: I would advocate to remove the dependence to pytest here, it's not required because we use the feature as a simple context manager. The boilerplate for pytest make that pretty heavy

```python
with setup_app_runner("./not_an_app", "run") as ar:
  pass
```

FabienArcellier:
```python
@contextlib.contextmanager
def setup_app_runner(app_dir: str, app_command: str):
    ar = AppRunner(app_dir, app_command)
    try:
        yield ar
    finally:
        ar.shut_down()
```

FabienArcellier:
```python
from conftest import setup_app_runner

def test_init_wrong_path(self) -> None:
    with setup_app_runner("./not_an_app", "run") as ar:
        with pytest.raises(SystemExit) as wrapped_e:
            ar.load()
        assert wrapped_e.type == SystemExit
        ```

PR #243, opened 2024-02-15
feat: added pdf viewer component

raaymax:
pdf viewer component with fallback to link download


ui/src/core_components/CorePDF.vue
```diff
@@ -0,0 +1,77 @@
...
+	"A component to embed a PDF document.";
+
+export default {
+	streamsync: {
+		name: "PDF",
+		description,
+		category: "Embed",
+		fields: {
+			source: {
+				name: "Pdf source",
```
ramedina86, line 27: Sorry for being a bit too intense with this stuff but can you capitalise PDF here?

ui/src/core_components/CorePDF.vue
```diff
@@ -0,0 +1,77 @@
...
+				type: FieldType.Text,
+				desc: "Either URL or Base64",
```
ramedina86, line 29: Please replace `desc` for

"A valid URL. Alternatively, you can provide a state reference to a packed PDF file."

The thing is, you cannot _strictly speaking_ use Base64 but rather a data URL which is a valid URL, which uses Base64 but isn't just Base64. Also, I think the `desc` is a good place to remind people that they can use packed files which are serialised to data URLs (in line with the `desc` in the _Image_ component).

ui/src/core_components/CorePDF.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+	<div class="CorePDF">
+		<object class="pdf" :data="fields.source.value" type="application/pdf" width="100%" height="100%">
```
ramedina86, line 3: Are these `width` and `height` properties necessary? They remind me of prehistoric HTML, haven't seen them in a while. I'm not familiar with `object` so maybe it's some strange thing that only applies to `object`?

ramedina86 (commented):
Just _really_ minor comments

PR #244, opened 2024-02-15
ci: added eslint checks for ui

raaymax:
On previous PR we lost some time fixing issues that should be handled by linter.
I checked and noticed that eslint was not working in the project so i fixed it.
Additionally added GH actions workflow for lint checking


ramedina86:
Sounds great, @FabienArcellier thoughts?

@raaymax What were the issues from the previous PR that the linter would have caught? (just curious)

raaymax:
Now as I review this PR, there isn't really much there. Actually, just console.log. Anyway, in the meantime, I fixed more things on my own, which is why it seemed like there was more.

raaymax:
From all this changes I only added missing `:key` bindings in `v-for` directives and some ignore comments. The rest was made by `eslint . --fix`

.github/workflows/ci-ui.yml
```diff
@@ -0,0 +1,17 @@
+name: ci-ui
```
FabienArcellier, line 1: could you add in `alfred ci.ui` instead of using github action for that ? it will run in local : https://github.com/streamsync-cloud/streamsync/blob/dev/alfred/ci.py and in the ci pipeline also.

alfred documentation : https://alfred-cli.readthedocs.io/en/latest/command.html

ramedina86:
Hey @raaymax are we good to merge this?

raaymax:
If @FabienArcellier doesn't have any more comments, then yes.

PR #248, opened 2024-02-19
chore: linking ComponentManager to session

mmikita95:
Making ComponentManager a part of session as a prerequisite to link the state of the UI to each particular session instead of the whole app.

### DIFF
--- a/src/streamsync/app_runner.py
+++ b/src/streamsync/app_runner.py
@@ -18,7 +18,7 @@ from typing import Any, Callable, Dict, List, Optional, cast
 from watchdog.observers.polling import PollingObserver
 
 from pydantic import ValidationError
-from streamsync.core import StreamsyncSession
+from streamsync.core import ComponentManager, StreamsyncSession
 from streamsync.ss_types import (AppProcessServerRequest, AppProcessServerRequestPacket, AppProcessServerResponse, AppProcessServerResponsePacket, ComponentUpdateRequest, ComponentUpdateRequestPayload,
                                  EventRequest, EventResponsePayload, InitSessionRequest, InitSessionRequestPayload, InitSessionResponse, InitSessionResponsePayload, StateEnquiryRequest, StateEnquiryResponsePayload, StreamsyncEvent)
 import watchdog.observers
@@ -134,7 +134,7 @@ class AppProcess(multiprocessing.Process):
         import traceback as tb
 
         session = streamsync.session_manager.get_new_session(
-            payload.cookies, payload.headers, payload.proposedSessionId)
+            payload.cookies, payload.headers, payload.proposedSessionId, self.components)
         if session is None:
             raise MessageHandlingException("Session rejected.")
 
@@ -149,7 +149,7 @@ class AppProcess(multiprocessing.Process):
             userState=user_state,
             sessionId=session.session_id,
             mail=session.session_state.mail,
-            components=streamsync.component_manager.to_dict(),
+            components=session.component_manager.to_dict(),
             userFunctions=self._get_user_functions()
         )
 
@@ -208,10 +208,6 @@ class AppProcess(multiprocessing.Process):
 
         return res_payload
 
-    def _handle_component_update(self, payload: ComponentUpdateRequestPayload) -> None:
-        import streamsync
-        streamsync.component_manager.ingest(payload.components)
-
     def _handle_message(self, session_id: str, request: AppProcessServerRequest) -> AppProcessServerResponse:
         """
         Handles messages from the main process to the app's isolated process.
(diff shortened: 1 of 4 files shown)


mmikita95:
Whoah, checks don't look good 😢 
Not something I was expecting, I'll look into it.

src/streamsync/core.py
```diff
@@ -723,7 +724,8 @@ class Evaluator:
 
     template_regex = re.compile(r"[\\]?@{([\w\s.]*)}")
 
-    def __init__(self, session_state: StreamsyncState):
+    def __init__(self, session_id: str, session_state: StreamsyncState):
+        self.session_id = session_id
```
ramedina86, line 728: Why are we passing an id instead of following what was done with `session_state` and passing component like `session_components`?

mmikita95: The id is currently needed to access component manager inside session(s) – previously, Evaluator used "application-wide" ComponentManager in the `evaluate_field` method; to replace it with session-linked manager, we first need to know what session we're working on specifically. I might've misunderstood the purpose of the StreamsyncState, but I couldn't see a way to utilize it for this.

src/streamsync/core.py
```diff
@@ -879,6 +885,13 @@ def __init__(self, session_id: str, cookies: Optional[Dict[str, str]], headers:
         new_state.user_state.mutated = set()
         self.session_state = new_state
         self.event_handler = EventHandler(self)
+        self.component_manager = ComponentManager()
```
FabienArcellier, line 888: I am not convince about the pattern integrating a manager in an instanced class.

I would advocate to keep the list of components the ComponentManager but we would ingest them with the session id. The component manager would stay a high level class we can rely on.

I would make the signature of ComponentManager evolve to keep the list of component per session id. It would be more easier to inspect application memory model and to make it evolve. We know all the components from all the session are in the ComponentManager. It may also allow advanced optimization as storing the diff between static component declaration and dynamic declaration.

We would also keep the ability to query the ComponentManager application wide. It may be interesting for some pattern.

src/streamsync/core.py
```diff
@@ -1075,7 +1088,7 @@ def handle(self, ev: StreamsyncEvent) -> StreamsyncEventResult:
         try:
             instance_path = ev.instancePath
             target_id = instance_path[-1]["componentId"]
-            target_component = component_manager.components[target_id]
+            target_component = self.session.component_manager.components[target_id]
```
FabienArcellier, line 1091: The code would look like this.

```python
component_manager.get_components(target_id, self.session.session_id)
```

src/streamsync/core.py
```diff
@@ -928,7 +941,7 @@ def _check_proposed_session_id(self, proposed_session_id: Optional[str]) -> bool
             return True
         return False
 
-    def get_new_session(self, cookies: Optional[Dict] = None, headers: Optional[Dict] = None, proposed_session_id: Optional[str] = None) -> Optional[StreamsyncSession]:
+    def get_new_session(self, cookies: Optional[Dict] = None, headers: Optional[Dict] = None, proposed_session_id: Optional[str] = None, components: Optional[Dict[str, Any]] = None) -> Optional[StreamsyncSession]:
```
FabienArcellier, line 944: We won't have to give the list of component instanciating the session.

src/streamsync/core.py
```diff
@@ -744,49 +746,53 @@ def replacer(matched):
...
+                if isinstance(repeater_object, dict):
+                    repeater_items = list(repeater_object.items())
+                elif isinstance(repeater_object, list):
+                    repeater_items = [(k, v)
+                                    for (k, v) in enumerate(repeater_object)]
+                else:
+                    raise ValueError(
+                        "Cannot produce context. Repeater object must evaluate to a dictionary.")
 
-        for i in range(len(instance_path)):
```
FabienArcellier, line 760: the code stay almost the same except on line 764

```
component_manager.get_components(component_id, self.session_id)
```

src/streamsync/app_runner.py
```diff
@@ -336,14 +332,14 @@ def _main(self) -> None:
             if self.mode == "run":
                 terminate_early = True
 
-        try:
-            streamsync.component_manager.ingest(self.components)
```
FabienArcellier, line 340:
```python
streamsync.component_manager.ingest(self.components, self.session_id)
```

FabienArcellier (commented):
I would advocate to keep the ComponentManager as high level class and to store inside the list of components associate with a session id.

src/streamsync/core.py
ramedina86: I think it's aligned with the current architecture. We're already managing sessions for State, my idea is to just put components there so we can manage them in the same way. There's a mechanism for example to clear out expired sessions and we'd have to replicate it for components.

src/streamsync/core.py
ramedina86: I agree that this shouldn't happen, should we mirrored in the same way that we handle initial state.

src/streamsync/core.py
ramedina86: But what I mean is, follow exactly what has been done with State. Rather than getting the session by id and then getting the component manager, pass the reference to the component manager when instantiating the Evaluator, similarly to how we pass `session_state`.

mmikita95: Ah, I see, I've misunderstood your comment, apologies. 

Good point, as session context is accessible during the creation of Evaluator, we might as well just pass the manager, without having the overhead of resolving it each time. I'll adjust it.

PR #253, opened 2024-02-19
feat: allow running review app to validate pr from external contribution

FabienArcellier:
The integration of an app review mechanism deploys a PR on demand to test the code online or demonstrate it without requiring cloning the branch locally.

Peek 2024-02-19 18-05

I used the reviewapp mechanism from [scalingo](https://scalingo.com/), a French PaaS platform. Their implementation is close to Heroku. After a first reading, project administrators will be able to trigger an environment to try and experiment on the code of the PR.

Peek 2024-02-19 17-24

The validation environment can be modified by any visitor. If this mechanism poses a problem, we will add authentication to limit access and only offer read access for visitors.

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,0 +1,2 @@
+https://github.com/FabienArcellier/nodejs-buildpack#streamsync-review
+https://github.com/FabienArcellier/python-buildpack#streamsync-review
--- a/apps/reviewapp.py
+++ b/apps/reviewapp.py
@@ -1,0 +1,43 @@
+import os
+
+import uvicorn
+from fastapi.responses import HTMLResponse
+from fastapi import FastAPI, Response
+
+import streamsync.serve
+
+
+def app_path(app_name: str) -> str:
+    return os.path.join(os.path.dirname(__file__), app_name)
+
+root_asgi_app = FastAPI(lifespan=streamsync.serve.lifespan)
+sub_asgi_app_1 = streamsync.serve.get_asgi_app(app_path("hello"), "edit", enable_remote_edit=True)
+sub_asgi_app_2 = streamsync.serve.get_asgi_app(app_path("default"), "edit", enable_remote_edit=True)
+sub_asgi_app_3 = streamsync.serve.get_asgi_app(app_path("quickstart"), "edit", enable_remote_edit=True)
+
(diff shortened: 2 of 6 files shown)


apps/reviewapp.py
```diff
@@ -0,0 +1,40 @@
+import os
```
FabienArcellier, line 1: Add a header to this module to explain why we have created the reviewapp.

.buildpacks
```diff
@@ -0,0 +1,2 @@
+https://github.com/FabienArcellier/nodejs-buildpack#streamsync-review
```
FabienArcellier, line 1: I don't know where to document the role of the Procfile and .buildpacks. ``.buildpacks`` does not support code comment.

ramedina86:
> The validation environment can be modified by any visitor. If this mechanism poses a problem, we will add authentication to limit access and only offer read access for visitors.

Is it too hard to implement auth with something like Google Auth and whitelist a few people?

Running `streamsync run` for everyone and `streamsync edit` for whitelisted would be ideal, but I'm sure harder to implement.

We'll be deploying many features that'll require us to handle API keys and maybe other sorts of confidential information. Being overly cautious but I'm scared someone could install a keylogger / etc, dump memory, or do something nasty.

pyproject.toml
```diff
@@ -34,6 +34,7 @@ watchdog = ">= 3.0.0, < 4"
 pandas = {version = ">= 2.2.0, < 3", optional = true}
 pyarrow = {version = ">= 15.0.0, < 16.0.0",optional = true}
 plotly = {version = ">= 5.18.0, < 6", optional = true}
+scikit-learn = {version = "^1.4.1.post1", optional = true}
```
ramedina86, line 37: Why is this here? Do we require it now for any reason?

FabienArcellier: We are using it in the getting started application.

ramedina86:
Hey @FabienArcellier what's the status of this? Are we going for the custom buildpack? If it's too hard we can even drop the quickstart app, I wouldn't spend too much time / add too much complexity to support this

FabienArcellier:
I have finished the work on the buildpack. This branch is not required anymore. I close it.

Issue #247, opened 2024-02-17
Event context not working for repeater objects where the repeater object is specified using square brackets "[ ]"


gunnar-enerin:
When using repeaters where the repeater object is specified using square brackets, by for example 
`@{example_state_object[sub_object]}`, Evaluator.evaluate_field() in core.py raises an error because the string passed to `json.loads(replaced)` does not contain valid json. This is caused by the regular expression not recognising the square brackets.

The use case where specifying the repeater object using square brackets is for instance with nested repeaters where a child repeater object needs to be specified by either the key or value of a parent repeater object.

This can be fixed by changing the regex line in the Evaluator class in core.py from:
`template_regex = re.compile(r"[\\]?@{([\w\s.]*)}")`
to
`template_regex = re.compile(r"[\\]?@{([\w\s.\[\]]*)}")`
to allow the recognition of square brackets.

I have tested that this fix actually resolves the issue, but I don't know whether it introduces any risks or other side effects.

mmikita95:
Thanks for spotlighting this issue and offering a solution — it is most appreciated!

Upon examination, your proposed fix seems clear of risks or side effects. The square brackets are indeed anticipated within the expression and are processed accordingly by `parse_expression` later on.

I encourage you to submit a PR to highlight your contribution. However, if you'd prefer not to, I'm more than willing to step in and do it on your behalf, just let me know!

mmikita95:
#257 is merged into dev branch, so we can consider this fixed!

PR #257, opened 2024-02-22
Add square barckets to evaluator regex

gunnar-enerin:
Resolves issue #247 

A minor fix to enable the possibility of having event contexts passed through nested repeater objects where the child repeater object is the item of a parent repeater.

### DIFF
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -743,7 +743,7 @@ class Evaluator:
     It allows for the sanitisation of frontend inputs.
     """
 
-    template_regex = re.compile(r"[\\]?@{([\w\s.]*)}")
+    template_regex = re.compile(r"[\\]?@{([\w\s.\[\]]*)}")
 
     def __init__(self, session_state: StreamsyncState, session_component_tree: ComponentTree):
         self.ss = session_state


ramedina86:
Thanks @gunnar-enerin for your contribution!

@FabienArcellier I think you've looked into this, do you have any input?

ramedina86:
Hey @mmikita95 since you already looked into this, can you provide your input? Also, let me know if you think we require any changes to our tests to avoid running into the issue that op detected.

mmikita95:
@ramedina86 good point about 
tests, I think we would definitely need it here. 

@gunnar-enerin could you please share your case that was triggering an error with a little more detail? I tried to reproduce it locally based on your description (my bad not doing it the first time and relying on code inspection solely 😅), and it avoids me for some reason – maybe I misunderstood something.

gunnar-enerin:
Hi, I should probably have added a concrete example initially as well, but I didn't have much time and wanted to get it out there. Sorry if that has caused extra work for you guys. 

It turns out (obviously in retrospect) that it has nothing in particular to do with nested repeaters, only using square brackets in the specification of the repeater object. Here is a minimal example that recreates the bug: https://github.com/gunnar-enerin/square-brackets-streamsync-bug-example.git 

The example is just a repeater object containing a section with a button inside where the repeater object is specified using square brackets notation (necessary when using dynamic specification, for instance when using nested repeaters, but not actually necessary in this case). Pressing the button causes a runtime error due to the regex not allowing square brackets.

ramedina86:
We've looked into this in detail @gunnar-enerin 

It's indeed a much needed fix. Thanks for spotting the issue and providing the fix!

It's quite impressive that you managed to get into the depths of Streamsync.

Hope to hear from you in the future!

PR #259, opened 2024-02-22
feat: updated pdf component with highlights

raaymax:


ui/src/core_components/embed/CorePDF.vue
```diff
@@ -33,26 +60,251 @@ export default {
...
+				default: null,
+				desc: "The index of the selected highlight match.",
+				type: FieldType.Number,
+			},
+			page: {
+				name: "Page",
+				type: FieldType.Number,
+				desc: "The page to be displayed.",
+			},
 			cssClasses,
```
ramedina86, line 80: Please also add separatorColor, backgroundColor and primaryTextColor so that users can configure these.

ui/src/core_components/embed/CorePDF.vue
```diff
@@ -33,26 +60,251 @@ export default {
...
 }
 
+.CorePDF .controls {
+	flex: 0 0 40px;
+	display: flex;
+	flex-direction: row;
+	justify-content: space-between;
+	align-items: center;
+	background-color: var(--backgroundColor);
+	border-bottom: 1px solid var(--separatorColor);
```
ramedina86, line 262: Please also add

`color: var(--primaryTextColor);`

For the color to be reflected (should happen in most cases but wouldn't rely on it)

ui/src/core_components/embed/CorePDF.vue
```diff
@@ -33,26 +60,251 @@ export default {
...
+	}
+	return offsetTop;
+};
+
+const gotoHighlight = (matchIdx: number) => {
+	if (matchIdx < 1 || matchIdx > matches.value.length) {
+		return;
+	}
+	const match = matches.value[matchIdx - 1];
+	console.log(match);
```
ramedina86, line 173: Please remove

ui/src/core_components/embed/CorePDF.vue
```diff
@@ -33,26 +60,251 @@ export default {
...
+			},
+			page: {
+				name: "Page",
+				type: FieldType.Number,
+				desc: "The page to be displayed.",
+			},
 			cssClasses,
 		},
+		events: {
+			"pdf-load": {
```
ramedina86, line 83: Is this event being fired? I can't seem to find the code that fires it.

Feel free to remove this event for now, but if you want to keep it:

- Make sure to trigger it
- Please rename this one to `ss-pdf-load`. You shouldn't need to change the core if there's no payload.
- Also, add a `desc` here even if we suspect what is, just for consistency.

**Note about "ss-" usage:**

I strongly prefer to prefix "ss-" and use sanitisation when needed, but when basically forwarding events from other libraries e.g. Google Maps this is potentially a pain as you usually have complex types that can change across versions.

ui/src/core_components/embed/CorePDF.vue
ramedina86: Also, play around with the colours and see how it goes, should render a decent image regardless of using default colours or not.

ui/src/core_components/embed/CorePDF.vue
```diff
@@ -33,26 +60,251 @@ export default {
...
+	border-bottom: 3px solid var(--separatorColor);
+	overflow: hidden;
+}
+
+.CorePDF .viewer .page > div {
+	width: fit-content;
+	margin: 0 auto;
+}
+
 .CorePDF.beingEdited:not(.selected) object {
```
ramedina86, line 308: Still needed?

ui/src/core_components/embed/CorePDF.vue
```diff
@@ -33,26 +60,251 @@ export default {
...
+	width: 40px;
+	line-height: 40px;
+	text-align: center;
+	border-radius: 4px;
+	flex: 0 0 40px;
+}
+
+.CorePDF .controls i:hover {
+	cursor: pointer;
+	background: var(--builderSubtleSeparatorColor);
```
ramedina86, line 284: The --builder vars are reserved for builder and won't work during runtime.

ui/src/core_components/embed/CorePDF.vue
```diff
@@ -1,17 +1,44 @@
...
-		>
-			<p>
-				You're not authorised to embed in the browser or you don't have
-				PDF plugin
-				<a :href="fields.source.value">download the PDF file. </a>
-			</p>
-		</object>
+	<div ref="rootEl" class="CorePDF">
+		<div class="controls">
+			<i class="ri-arrow-up-s-line" @click="() => gotoPage(page - 1)"></i>
```
ramedina86, line 4: For accessibility / semantic HTML purposes I try to use `button` when something is a button. Also, makes them keyboard navigable. Can you please put the `i`s in `button`s?

ui/src/core_components/embed/CorePDF.vue
ramedina86: Also, I once remember reading about some guy having a linting rule to get rid of `console.log` messages. Since you're good at that stuff maybe you can implement something similar?

I'd still want to keep the loading messages (legitimate uses of `console.log`) but that should probably be handled in a more professional way, defining a logger and then invoking that logger.

ui/src/core_components/embed/CorePDF.vue
```diff
@@ -33,26 +60,251 @@ export default {
...
+				name: "Selected highlight match",
+				default: null,
+				desc: "The index of the selected highlight match.",
+				type: FieldType.Number,
+			},
+			page: {
+				name: "Page",
+				type: FieldType.Number,
+				desc: "The page to be displayed.",
+			},
```
ramedina86, line 79: Can we add a toggle controls "yes / no" as discussed?

ui/src/core_components/embed/CorePDF.vue
raaymax: 👍 
Rule added to eslint
you can use:
```js
// eslint-disable-next-line no-console
```
if you need to keep logging

ui/src/core_components/embed/CorePDF.vue
raaymax: 👍

ui/src/core_components/embed/CorePDF.vue
ramedina86: Great stuff

ramedina86:
Thanks for the changes and the fix

PR #260, opened 2024-02-23
fix: added map type options to google maps component

raaymax:


ramedina86:
Thanks, makes configuration easier and less error prone.

PR #261, opened 2024-02-23
refactor: component base with built-in mask

raaymax:
Proposition for introduction of `ComponentBase` with built-in mask.


ramedina86:
Hey sorry didn't expect you to actually work on this, I actually just wanted you to get your thoughts on applying a mask to everything. I should have been more explicit. Let's go through this together, if we do decide to apply I believe it should be in `ComponentProxy` through which all components are rendered.

ramedina86:
@raaymax As discussed, I got my head around it, I think it's a great idea if we follow a few points:

- We change the name to `BaseMaskedComponent` (or similar)
- We use exclusively in those components that require it
- We document in the code why masks are needed (I think it'd help the team and also custom component creators) 

A `ComponentProxy` mask is nice because of its "set and forget" nature, but does add complexity. After looking at the `ComponentProxy` code, I think it'd be quite dirty as it'd also require injecting a node inside the parent node, rather than wrapping the node. Feel free to have a look if interested and see if you can come up with something more elegant, but I think we're good with this one.

raaymax:
Changes you mentioned are now applied.
Additionally I added this mask to video player component to fix it on firefox.

ramedina86:
@raaymax What should we do about this one? Are we still going to use it?

PR #266, opened 2024-02-26
feat: added link component

raaymax:
This needs double check, I added dynamic list of options for components.


ui/src/builder/BuilderFieldsText.vue
```diff
@@ -76,6 +76,16 @@ const templateField = computed(() => {
 	return definition.fields[fieldKey.value];
 });
 
+const options = computed(() => {
+	const field = templateField.value;
+	if (field.options) {
+		return typeof field.options === "function"
```
ramedina86, line 82: Hey @FabienArcellier check this out

@raaymax worked on a dynamic list of options. You can pass it hardcoded options (as we do now), but you can also pass it a function to generate the options.

This was one of the challenges I saw for the _Reuse Component_ component. So with this we're halfway there. I've asked him to work on _Reuse Component_.

@raaymax for your reference https://github.com/streamsync-cloud/streamsync/issues/215

ui/src/core_components/content/CoreLink.vue
```diff
@@ -0,0 +1,86 @@
...
+	},
+};
+</script>
+
+<script setup lang="ts">
+import { inject, computed } from "vue";
+const ss = inject(injectionKeys.core);
+const fields = inject(injectionKeys.evaluatedFields);
+
+options = computed(() => {
```
ramedina86, line 66: Can you fix this with `const options`? You've defined is as `[]` above but as `ComputedRef` here.

ui/src/core_components/content/CoreLink.vue
```diff
@@ -0,0 +1,86 @@
...
+	return ss
+		.getComponents("root", true)
+		.map((page) => page.content.key)
+		.filter((key) => Boolean(key))
+		.map((key) => [`#${key}`, key]);
+});
+
+const displayText = computed(() => {
+	return fields.text.value || fields.url.value || "Link";
```
ramedina86, line 75: Can you use a `default` in the field, rather than ` || "Link"`? Should achieve the same and is more consistent with the rest of the components.

ui/src/core_components/content/CoreLink.vue
```diff
@@ -0,0 +1,86 @@
...
+});
+</script>
+
+<style scoped>
+.CoreLink a {
+	color: var(--primaryTextColor);
+}
+.CoreLink.beingEdited:not(.selected) a {
```
ramedina86, line 83: I think it should work without this. Can you try? I think you're traumatised after working with PDFs etc. But under normal circumstances the existing code should take care of intercepting the event.

ui/src/core_components/content/CoreLink.vue
```diff
@@ -0,0 +1,86 @@
...
+
+<script setup lang="ts">
+import { inject, computed } from "vue";
+const ss = inject(injectionKeys.core);
+const fields = inject(injectionKeys.evaluatedFields);
+
+options = computed(() => {
+	return ss
+		.getComponents("root", true)
+		.map((page) => page.content.key)
```
ramedina86, line 69: Keep in mind `key` is optional and most likely empty.

ui/src/core_components/content/CoreLink.vue
```diff
@@ -0,0 +1,86 @@
...
+export default {
+	streamsync: {
+		name: "Link",
+		description: "A component to create a hyperlink.",
+		category: "Content",
+		fields: {
+			url: {
+				name: "URL",
+				type: FieldType.Text,
+				desc: "A valid URL.",
```
ramedina86, line 28: I was quite confused by the list of pages at first, when I got to try this component. Took me a while to be like "Ah, of course, the pages are accessible via #".

I love the new dynamic options capability, but it makes you think the right options are the ones listed, when I expect most people to enter a "standard" URL e.g. "github.com". 

If we keep it:
- I'd make it clearer in the `desc` like, "Specify a URL or choose a page."
- I'd set the default to `https://streamsync.cloud`. Bit of self promotion but it'd make it clear right away that yes, you can include "normal" URLs and not just #fragments.

I'm also not sure how it'd behave when the list is empty, needs checking. I expect most apps to not have any keyed pages.

---

Also, something to keep in mind... The hash can contain parameters that will be lost when switching pages this way.

https://www.streamsync.cloud/page-routes.html#routes-with-parameters

For example if you have `/#detailPage/product_id=32&country=AR` and you switch via event handler or [backend action](https://www.streamsync.cloud/backend-initiated-actions.html#changing-the-active-page) it'll preserve the rest of the hash.

If someone types `#mypage` then they literally asked for the parameters to be removed, but if they select it from a list they might have the expectation that it behaves like other page-jumping functionality.

---

I'm more inclined towards not keeping the dynamic options here, but they'd certainly come in handy for _Reuse Component_.

ramedina86: The logic that manages the hash is in `CoreRoot.vue` if you're interested.

ui/src/builder/BuilderFieldsText.vue
ramedina86: By the way I'm talking specifically about selecting the `id` of the component that we want to reuse.

ui/src/core_components/content/CoreLink.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+	<div class="CoreLink">
```
ramedina86, line 2: Since you're working on links, can you please check how the Markdown links are working, if everything is alright with them?

I assume that with a _Text component_, markdown mode, I can `[hello](https://streamsync.cloud)`.

ui/src/core_components/content/CoreLink.vue
ramedina86: I see that we've been explicitly asked to link pages. So maybe let's try how it'd look like with the minor changes I suggested under "If we keep it".

`desc` should probably be "Specify a URL or choose a page. Keep in mind that you can only link to pages for which a key has been specified."

Keen to hear your thoughts. Apologies for the verbosity, you probably weren't expecting an essay after submitting a PR for an `a href`.

ui/src/builder/BuilderFieldsText.vue
raaymax: I investigated this case, and it seems that it won't be that simple. This specific example works, but unfortunately, this approach won't work for most cases. Actually, this isn't valid code from the perspective of this framework, unfortunately. I need to investigate it further; maybe I will find a better solution for that.

ui/src/core_components/content/CoreLink.vue
raaymax: Without it I can't select link, it immediately redirects to target location.

ui/src/core_components/content/CoreLink.vue
```diff
@@ -0,0 +1,83 @@
...
+		fields: {
+			url: {
+				name: "URL",
+				type: FieldType.Text,
+				desc: "A valid URL.",
+				options: (ss: Core) => {
+					return Object.fromEntries(
+						ss
+							.getComponents("root", true)
+							.map((page) => page.content.key)
```
ramedina86, line 31: Careful with this one. Like I was saying earlier not all pages will have keys. In fact I expect most pages to not have keys.

ramedina86: Ah sorry didn't see the next line

ui/src/core_components/content/CoreLink.vue
ramedina86: Thoughts on this one?

ui/src/core_components/content/CoreLink.vue
raaymax: thats why there is filter in the next line

ui/src/core_components/content/CoreLink.vue
ramedina86: Makes sense. 

I had an idea, I think we won't need masks at all. In `BuilderApp.vue` go to `handleRendererClick` and above the `ev.stopPropagation()` add `ev.preventDefault()`. Worked for `a` made with _HTML Element_.

ui/src/core_components/content/CoreLink.vue
raaymax: I think it's just how these kinds of links work. Personally, it would be really confusing for me if the link appended parameters at the end of the URL. I expect that most people using this app will have at least a really basic knowledge about HTML, and <a href> is probably the first thing they learn.

ui/src/core_components/content/CoreLink.vue
raaymax: Just checked and yes - you can. You can also `[Next page](#next-page)` to navigate between pages.

raaymax: So it's basically the same, only difference is that you can select page in component and define rel and target.

ramedina86: Thanks for checking!

Yes they're quite similar, that's what kept me from creating this component. However, now that I think more about it:
- Agreed, there's the `rel` and `target` 
- You can change the color
- It's more semantic. If you see a component tree with 3 _Link_, you it's self-explanatory. You see 3 _Text_, could be anything.
- It's more easily discoverable. "I need to create a link, so I'll create a _Text_ component which probably supports markdown and I'll `[]()` my way into it." -Probably no one ever, at least not on their first try.

By the way there's the possibility of opening URLs from the backend via backend actions too.

PR #268, opened 2024-02-27
tests: frontend test framework setup with playwright

raaymax:


src/streamsync/serve.py
```diff
@@ -408,7 +408,7 @@ def print_route_message(run_name: str, port: int, host: str):
     GREEN_TOKEN = "\033[92m"
     END_TOKEN = "\033[0m"
 
-    print(f"{run_name} is available at:{END_TOKEN}{GREEN_TOKEN} http://{host}:{port}{END_TOKEN}")
+    print(f"{run_name} is available at:{END_TOKEN}{GREEN_TOKEN} http://{host}:{port}{END_TOKEN}", flush=True)
```
ramedina86, line 411: Really curious about the `flush`

ui/src/builder/BuilderComponentShortcuts.vue
```diff
@@ -109,6 +109,7 @@
 			</div>
 			<div
 				class="actionButton delete"
+				data-action="delete"
```
ramedina86, line 112: Any chance we can use some sort of namespace like `data-automation-action` or is that not configurable?

ui/tests/button.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	test.beforeAll(async() => {
+		await loadPreset('base');
+	});
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto("/");
+		test.setTimeout(5000);
+	});
+
```
ramedina86, line 19: Amazing, it's beautiful 🥹

ui_test_app/index.js
```diff
@@ -0,0 +1,88 @@
...
+		await this.start();
+	}
+}
+
+const ss = new Streamsync();
+(async () => {
+	await fs.mkdir("runtime", { recursive: true });
+	await ss.loadPreset("base");
+})();
+const app = express();
```
ramedina86, line 77: Is this the magic you were talking about? 🧙‍♂️

src/streamsync/serve.py
raaymax: Thats needed by `ui_test_app` which is determining readiness status of the sub process based on this message. Sometimes program was stuck on waiting for this message when it was buffered.

ui_test_app/index.js
raaymax: yup, It's a wrapper for Streamsync for reloading ui configuration file. Tests live on the frontend side so i needed some way to communicate with it to be able to trigger reload.

ui/src/builder/BuilderComponentShortcuts.vue
raaymax: Good idea! We can name it however we want so I will change this.

e2e_tests/tests/createAndRemove.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+	{ type: "selectinput", locator: `div.CoreSelectInput.component` },
+	{ type: "multiselectinput", locator: `div.CoreMultiselectInput.component` },
+	{ type: "fileinput", locator: `div.CoreFileInput.component` },
+	{ type: "webcamcapture", locator: `div.CoreWebcamCapture.component` },
+	{ type: "vegalitechart", locator: `div.CoreVegaLiteChart.component` },
+	{ type: "plotlygraph", locator: `div.CorePlotlyGraph.component` },
+	{ type: "metric", locator: `div.CoreMetric.component` },
+	{ type: "message", locator: `div.CoreMessage.component` },
+	{ type: "videoplayer", locator: `div.CoreVideoPlayer.component` },
+];
```
ramedina86, line 48: I have an idea, we should do a pass of the whole `templateMap` and check whether we have tests for everything mentioned there. If we don't, we fail.

Otherwise we can easily forget.

Or maybe some `--strict` flag so that we can pass a PR but not release?

Keen to hear your thoughts.

e2e_tests/tests/createAndRemove.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+const fullCheck = [
+	{ type: "button", locator: `button.CoreButton.component` },
+	{ type: "text", locator: `div.CoreText.component` },
+	{ type: "header", locator: `div.CoreHeader.component` },
+	{ type: "heading", locator: `div.CoreHeading.component` },
+	{ type: "dataframe", locator: `div.CoreDataframe.component` },
+	{ type: "html", locator: `div.CoreHTML.component` },
+	{ type: "pagination", locator: `div.pagination.component` },
+	{ type: "repeater", locator: `div.CoreRepeater.component` },
+	//{ type: "column", locator: `div.CoreColumn.component` },
```
ramedina86, line 20: If you want to automate something really cool you could check the `allowedParentTypes` and automatically create the parents.

e.g. from `tab`:

`allowedParentTypes: ["tabs", "repeater"]`

So you could exclude `repeater` and you have the right parent.

That'd work for `column`, `step`, `tab`.

e2e_tests/tests/createAndRemove.spec.ts
raaymax: I was thinking about using `templateMap` here earlier but I don't have access to this file from here. Thats why I just created list of components base on `templateMap`.

e2e_tests/tests/createAndRemove.spec.ts
raaymax: This would be really nice but lets save it as future improvement, first need to stabilize what we already have.

e2e_tests/tests/createAndRemove.spec.ts
ramedina86: If you have access to the `core` you can use `getSupportedComponentTypes` without doing an import of `templateMap` which I agree, wouldn't be the cleanest.

ramedina86:
I'd really like to start benefiting from this soon, as it gives me more confidence not only when it comes to frontend but also end-to-end. There's the Low Code UI work going on which is big and disruptive, also Fabien's latest PR.

Any chance we can merge something in the next few days?

raaymax:
I'm doing what i can to resolve all issues before making it ready to review. 
Currently only chromium tests are passing firefox and safari still needs some work.
But I think positively and hope that everything will be resolved soon.

e2e_tests/index.js
```diff
@@ -0,0 +1,148 @@
...
+        ss.kill();
+        reject();
+      }, 5000);
+
+      ss.stdout.on("data", (data) => {
+        // eslint-disable-next-line no-console
+        console.log(
+          `[${ss.pid}] stdout: ${Buffer.from(data, "utf-8").toString()}`,
+        );
+        if (data.includes("Builder is available at")) {
```
ramedina86, line 36: Any ideas to make this less hacky in the future? Something like a health check, some signal I don't know about?

Probably useful for other purposes too.

e2e_tests/tests/undoRedo.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { test, expect } from "@playwright/test";
+
+test.setTimeout(5000);
+
+test.describe('undo and redo', () => {
```
ramedina86, line 5: Awesome that you included this

ramedina86 (approved):

Issue #277, opened 2024-03-02
declare optional schema on streamsync state


FabienArcellier:
The schema offer a way to associate a streamsync state with an explicit schema. The developer can specify how the state he builds in streamsync is structured.

This parameter is optional. If it is not provided, the current behavior which consists of interpreting dictionaries as substates is preserved.

This mechanism may solve 2 problems:

* specifies the type and allows plotly or vegas payloads not to be interpreted as a state (https://github.com/streamsync-cloud/streamsync/issues/226)
* offer autocompletion and type checking

# Simple example

```python
import streamsync as ss

class AppSchema(ss.StreamsyncState):
    counter: int

initial_state = ss.init_state({
    "counter": 0
}, schema=AppSchema)

# Event handler
# It receives the session state as an argument and mutates it
def increment(state: AppSchema):
    state.counter += 1
```

ramedina86:
Shipped in 0.4.0. Thanks @FabienArcellier !

Issue #171, opened 2023-11-09
Q: Event item - context in repeater component


joelwiesmann:
Hey!

First of all: thank you for this incredible piece of software! I'm sure this project will have a bright future!

I'm using a repeater component to display data. For each item I'd like to create a button that does some action on it. Unfortunately, the payload of the click-event on the button does not feature any repeater-item context. So every button in the list will basically do the same.

Is there a way to access the "itemId" or any other way to provide context / parameters to an event?

Regards
Joel

joelwiesmann:
Here's a visualization based on the example data. The button should remove the item from the dictionary. However right now, it doesn't know which item is being "eaten".
![image](https://github.com/streamsync-cloud/streamsync/assets/23297989/fbc2fa79-9683-4d8c-9e3b-195b9a8b205e)

FabienArcellier:
Hi, the item called in a repeater can be accessed through a context parameter in event handler. 
Did you try this ?

```python
def item_remove(state, context):
    items = state["items"].state
    items = {k: v for k, v in items.items() if k != context["item"]}
    state["items"] = items
```

joelwiesmann:
Hi @FabienArcellier 

Oh interesting, I wasn't aware of **context**, thanks for the hint!

I amended your example as I think context["item"] should have been context["itemId"]. Now something strange happens - the first item that is being removed properly works. When I click a second time, the state["items"] is null.

![10-11-2023_07-38-11](https://github.com/streamsync-cloud/streamsync/assets/23297989/77516860-fecf-4139-a6a1-3e98f18ce0e9)

ramedina86:
@mmikita95 something else to look into, possibly in line with the kind of things you've been focusing on.

mmikita95:
Having a discussion about this, we've decided to introduce a "removal" flag for keys that are carried to the frontend, so that we can track these kinds of changes in state more consistently and reliably. Basically, prepend the keys to be removed with '-', and the rest of them with '+'. This will not influence users' naming conventions, as we'll only track the first character for this flag, which will be added during the preparation of mutations. I'll be working on this on the backend's side now.

PR #278, opened 2024-03-02
feat:declare optional schema on streamsync state

FabienArcellier:
implement #277


src/streamsync/core.py
```diff
@@ -1125,10 +1250,68 @@ def handle(self, ev: StreamsyncEvent) -> StreamsyncEventResult:
...
+    Initializes the initial state of the application and makes it globally accessible.
+
+    The class used for the initial state must be a subclass of StreamsyncState.
+
+    >>> class MyState(StreamsyncState):
+    >>>     pass
+    >>>
+    >>> initial_state = new_initial_state(MyState)
+    """
+    global initial_state
```
FabienArcellier, line 1311: Rewriting a module variable there introduces a risk of introducing futur bugs with current accessors. When a variable is import from external module, its value is freezed. There is a risk to replace the value of the variable without updating external module.

To reduce the risk of issue, we should expose a function that allow to get the variable `initial_state()`. Using a function avoid the risk for this issue.

src/streamsync/core.py
```diff
@@ -204,6 +203,9 @@ def ingest(self, raw_state: Dict) -> None:
         for key, raw_value in raw_state.items():
             self.__setitem__(key, raw_value)
 
+    def items(self) -> Sequence[Tuple[str, Any]]:
```
ramedina86, line 206: What is this? Is this necessary or a separate improvement to make it look more like a Dict?

src/streamsync/core.py
ramedina86: So, keeping the global variable but using an accessor rather than importing it directly?

Like this?

```py
def get_initial_state():
  return initial_state
```

FabienArcellier: yes exactly

mmikita95 (commented):
_Potentially reintroduces_ #171; _clarification needed._

src/streamsync/core.py
```diff
@@ -215,19 +217,7 @@ def __setitem__(self, key, raw_value) -> None:
...
-        elif isinstance(raw_value, StateProxy):
-            # Children StateProxies need to be reinitialised
-            # during an assignment to parent StateProxy
-            # to ensure proper mutation tracking
-            value = StateProxy(raw_value.state)
-        else:
-            value = raw_value
-
-        self.state[key] = value
+        self.state[key] = raw_value
```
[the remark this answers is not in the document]
FabienArcellier: I have to move StateProxy creation in the State because, it's this object that know the schema. You should take a look on the method `State._set_state_item`.

tests/test_core.py
```diff
@@ -141,6 +132,102 @@ def test_private_members(self) -> None:
...
+        }})
+
+        # When
+        items = _state['items']
+        items = {k: v for k, v in items.items() if k != "Apple"}
+        _state["items"] = items
+
+        # Then
+        m = _state._state_proxy.get_mutations_as_dict()
+        assert m == {
```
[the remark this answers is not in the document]
FabienArcellier: Thanks @mmikita95. There was something wrong. I set some changes trying to fix it. Could you take a look ?

raaymax:
I see that tests for state propagation proven to be useful :D 
```
    Expected string: "{\"a\":1,\"b\":2,\"c\":{\"d\":3}}"
    Received string: "{\"a\":1,\"b\":2,\"c\":null}"
```

PR #279, opened 2024-03-05
feat!: core managed components functionality

mmikita95:
- Moving Component, ComponentTree and SessionComponentTree classes to a separate ui_utils file.
- StreamsyncUI class responsible for creating components and attaching them to ComponentTree
- StreamsyncUIManager class intended as a "user-facing" interface
- Tests for UI manager


src/streamsync/ui_manager.py
```diff
@@ -0,0 +1,70 @@
...
+            parent_id = kwargs.pop('parentId')
+        else:
+            parent_id = 'root' if not parent_container else parent_container.id
+        component = Component(
+            type=component_type, parentId=parent_id, flag='cmc', **kwargs
+            )
+        self.component_tree.attach(component, attach_to_bottom=True)
+        return component
+
+    def create_container(self, component_type: str, **kwargs) \
```
ramedina86, line 61: Can we please rename to `create_container_component`?

src/streamsync/ui_manager.py
```diff
@@ -0,0 +1,70 @@
...
+    Streamsync session.
+
+    This class offers context managers and methods to dynamically create, find,
+    and organize UI components based on a structured component tree.
+    """
+
+    def __init__(self, session_id: str):
+        self.session: Union[StreamsyncSession, None] = session_manager.get_session(session_id)
+        if self.session is None:
+            raise RuntimeError('Invalid session passed to the UI manager')
```
ramedina86, line 19: I'd prefer double quotes to be used consistently, but enforcing this via code reviews isn't good for anyone's sanity.

I'll create another issue to install `pylint`, which I'm sure will improve the quality of our code in other ways too.

src/streamsync/ui_utils.py
```diff
@@ -0,0 +1,141 @@
+from contextvars import ContextVar
+from typing import Any, Dict, List, Optional, Union
+import uuid
+
+from pydantic import BaseModel, Field
+
+current_parent_container: ContextVar[Union['Component', None]] = \
+    ContextVar('current_parent_container')
+# This variable is thread safe and context safe
```
ramedina86, line 9: Thanks for this comment, I was wondering this

src/streamsync/ui_utils.py
```diff
@@ -0,0 +1,141 @@
+from contextvars import ContextVar
```
FabienArcellier, line 1: I would advocate to rename this module `code_ui.py`. It contains datastructure that are useful for the rest of application. That's easy to communicate the core exist for internal purpose. We may also move into `core/ui.py` but it would mean we put the current core into `core/__init__.py.

ramedina86: I'd love a better name, but I can't think of one.

I don't think `code_ui` is descriptive as it's foundational code also used by BMC.

FabienArcellier: sorry `core_ui`

ramedina86: I get it now, and I fully agree with this recommendation.

FabienArcellier (changes requested):
Thanks, it's a great step forward. 

We have to cover this pattern :

```python
import streamsync as ss

with ss.init_ui() as ui:
  with ui.Page("other page"):
    ui.Text(text='Hello Pigeons')
```

tests/test_ui.py
```diff
@@ -0,0 +1,91 @@
...
+        assert container.id in self.ui_manager.session.session_component_tree.components.keys()
+
+    def test_column_creation(self):
+        column = self.ui_manager.Column(id="test_column")
+        assert column.type == 'column'
+        assert column.id == "test_column"
+        assert column in self.ui_manager.session.session_component_tree.components.values()
+
+    def test_text_creation(self):
+        with self.ui_manager.root:
```
FabienArcellier, line 84: I think `root` accept usually only pages in streamsync. I don't think it's a good way to start playing with user interface. What's happen when you set this in the json file ?

src/streamsync/ui_manager.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def root(self) -> Component:
+        if not self.session:
+            raise RuntimeError("No session object on UI manager")
+        if not self.root_component:
+            raise RuntimeError(f"Failed to acquire root component in session {self.session.session_id}")
+        return self.root_component
+
+    def find(self, component_id: str) \
```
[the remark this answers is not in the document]
mmikita95: It can work as one due to the return being a component, which supports context management. I did it so that users could also directly assign components to variables and modify properties, in line with what @ramedina86 was suggesting. If there's some extra benefits from adding this decorator that I'm overlooking, please let me know.

src/streamsync/ui_manager.py
```diff
@@ -0,0 +1,70 @@
...
+        if component is None:
+            raise RuntimeError(f"Component {component_id} not found")
+        return component
+
+    def _create_component(self, component_type: str, **kwargs) -> Component:
+        parent_container = current_parent_container.get(None)
+        if "parentId" in kwargs:
+            parent_id = kwargs.pop("parentId")
+        else:
+            parent_id = "root" if not parent_container else parent_container.id
```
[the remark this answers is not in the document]
mmikita95: Yeah, this is the catch: I've had to allow to create "container" components (like Column and ColumnContainer) without `assert_in_container` check, because otherwise we wouldn't have a way to create it at all when we begin working with the manager. I think that when we've added the rest of the components, we might reserve this logic to pages – modifying it slightly, as it appears pages are always children only to root component.

tests/test_ui.py
```diff
@@ -0,0 +1,91 @@
+from streamsync.core_ui import ComponentTree, UIError
+from streamsync.ui import StreamsyncUIManager
+import streamsync as ss
+
+import json
+
+from tests import test_app_dir
+
+sc = None
```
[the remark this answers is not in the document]
mmikita95: Good point. I think that we're better off sticking to the first scenario, as it is more in line with how we've been designing other tests already.

src/streamsync/ui_manager.py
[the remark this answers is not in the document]
FabienArcellier: Sorry I didn't push the complete implementation on this one:

```python
@contextlib.contextmanager
def find(self, component_id: str) -> Generator[Component, None, None]:
    # Example context manager for finding components
    component = self.component_tree.get_component(component_id)
    if component is None:
        raise RuntimeError(f'Component {component_id} not found')
    yield component
```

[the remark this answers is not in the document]
FabienArcellier: Ok, I didn't know it was working without it.

src/streamsync/ui_manager.py
raaymax, line 1: question: is this intended?
class `StreamsyncUI` is in `ui_manager.py`
and class `StreamsyncUIManager` in in `ui.py`

mmikita95: I see how this can be confusing. My logic was that `ui_manager` contains an "actual manager class" (`StreamsyncUI`), and `ui` provides easy access to the interface (`StreamsyncUIManager`) that user is supposed to build upon. Hence the discrepancy in the class names and file names – `ui_manager` is reserved for internal use, and `ui` is "out-facing", while the class names are vice versa, so that `StreamsyncUIManager` could give a user a better hint on what it is supposed to do.
I think we might change that though, because, on contemplating on it, doesn't seem like the best choice.

src/streamsync/core.py
```diff
@@ -1079,6 +983,10 @@ def _call_handler_callable(self, event_type, target_component, instance_path, pa
                     "headers": self.session.headers
                 }
                 arg_values.append(session_info)
+            elif arg == "ui":
+                from streamsync.ui import StreamsyncUIManager
+                ui_manager = StreamsyncUIManager(self.session.session_id)
```
ramedina86, line 988: I'd pass the session tree as an argument of `StreamsyncUIManager` rather than the `session_id`.

Then you can create `StreamsyncUIManager` attached to the base tree when outside of a handler.

src/streamsync/ui_manager.py
ramedina86: I don't know why this is wrapped inside a class.

If users have to `StreamsyncUIManager.Text()` that's a pain, we expected `ui.Text()`. Is this not it?

Apologies I missed this earlier, I definitely didn't have a good enough look.

src/streamsync/core_ui.py
```diff
@@ -0,0 +1,141 @@
...
+            override=False,
+            attach_to_bottom=False) -> None:
+        if attach_to_bottom is True and component.position == 0:
+            if component.parentId is not None:
+                tree = self \
+                        if component.parentId in self.components \
+                        else self.base_component_tree
+                parent_length = \
+                    tree.get_direct_descendents_length(component.parentId)
+                component.position = parent_length + 1
```
ramedina86, line 124: To get the position you should check the maximum position, then add 1.

We need to account for positionless components (only _Sidebar_ at this point)

Creation of _Sidebar_ should also hardcode position to `-2`. This is a special value that indicates that it should be rendered in the positionless slot.

ramedina86: Actually, check out `getNextInsertionPosition()` in `useComponentActions.ts`.

If there are children, should be the 
Math.max(max position + 1, 0)

No children:
0

PR #283, opened 2024-03-06
fix: drop target highlighting is moved when the Settings bar is open

raaymax:


ui/src/builder/BuilderInstanceTracker.vue
```diff
@@ -60,15 +64,17 @@ const trackElement = (el: HTMLElement) => {
...
 	}
 
-	let correction = 0;
-	if (settingsLeft < rendererX + rendererWidth) {
-		const trackerEnd = trackerX + contentsWidth;
-		const rendererEnd = rendererX + rendererWidth;
-		const distanceToRight = Math.max(rendererEnd - trackerEnd, 0);
-		correction = Math.max(settingsWidth - distanceToRight, 0);
-	}
+	if (!props.ignoreSettingsSidebar) {
```
ramedina86, line 67: Two minor things that I think can help with readability and consistency:

- Either change everything to use `props.` or use the `toRefs`
- Change from a `!ignoreSettingsSidebar` logic to `preventSettingsBarOverlap`, and just apply it to the one that needs it.

PR #289, opened 2024-03-07
chore: (minor) adding dash to template_regex in Evaluator class

mmikita95:
Adding a dash to this regular expression enables support for "skewer-case" state property names on frontend side.

### DIFF
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -749,7 +749,7 @@ class Evaluator:
     It allows for the sanitisation of frontend inputs.
     """
 
-    template_regex = re.compile(r"[\\]?@{([\w\s.\[\]]*)}")
+    template_regex = re.compile(r"[\\]?@{([\w\s.\[\]]*)}", re.U)
 
     def __init__(self, session_state: StreamsyncState, session_component_tree: ComponentTree):
         self.ss = session_state
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -741,7 +741,6 @@ class EventDeserialiser:
             return None
 
 
-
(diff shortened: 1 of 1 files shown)


src/streamsync/core.py
```diff
@@ -749,7 +749,7 @@ class Evaluator:
     It allows for the sanitisation of frontend inputs.
     """
 
-    template_regex = re.compile(r"[\\]?@{([\w\s.\[\]]*)}")
+    template_regex = re.compile(r"[\\]?@{([\w\s.\[\]]*)}", re.U)
```
mmikita95: As I eventually found out, this flag is redundant and included for backward compatibility.

src/streamsync/core.py
```diff
@@ -909,15 +909,14 @@ def _transform_change_page(self, ev) -> Optional[int]:
...
-
 class Evaluator:
 
     """
     Evaluates templates and expressions in the backend.
     It allows for the sanitisation of frontend inputs.
     """
 
-    template_regex = re.compile(r"[\\]?@{([\w\s.\[\]]*)}")
+    template_regex = re.compile(r"[\\]?@{([^{]*)}")
```
mmikita95, line 919: `r"[\\]?@{([^{]*)}"`

`[\\]`: matches a single character present in the list (`\\` matches the character `\` literally)
`@{`: matches the characters `@{` literally
Capturing Group `([^{]*)`: matches a single character not present in the list (excludes `{`)
`*`: matches the previous token between zero and unlimited times, as many times as possible, giving back as needed (greedy)
`}`: matches the character `}` literally

PR #297, opened 2024-03-12
chore: add init_ui method

mmikita95:
Adding a wrapper `init_ui` method, to allow the syntax proposed by @FabienArcellier:
```
with ss.init_ui() as ui:
    ...
```

### DIFF
--- a/src/streamsync/__init__.py
+++ b/src/streamsync/__init__.py
@@ -1,6 +1,7 @@
 import importlib.metadata
 from typing import Union, Optional, Dict, Any
 from streamsync.core import Readable, FileWrapper, BytesWrapper, Config
+from streamsync.ui import StreamsyncUIManager
 from streamsync.core import initial_state, base_component_tree, session_manager, session_verifier
 
 VERSION = importlib.metadata.version("streamsync")
@@ -38,3 +39,10 @@ def init_state(state_dict: Dict[str, Any]):
     initial_state.user_state.state = {}
     initial_state.user_state.ingest(state_dict)
     return initial_state
+
+
+def init_ui():
+    """
+    Wrapper for initializing the UI manager.
+    """
+    return StreamsyncUIManager()


src/streamsync/__init__.py
```diff
@@ -38,3 +39,10 @@ def init_state(state_dict: Dict[str, Any]):
     initial_state.user_state.state = {}
     initial_state.user_state.ingest(state_dict)
     return initial_state
+
+
+def init_ui():
```
FabienArcellier, line 44: I would advocate to add typing information using Union of Generator (for mypy and vscode) and ContextManager (for pycharm)

src/streamsync/__init__.py
```diff
@@ -38,3 +39,10 @@ def init_state(state_dict: Dict[str, Any]):
...
+    """
+    Wrapper for initializing the UI manager.
+    """
```
FabienArcellier, line 47: Improve documentation, here it's for developper, not user

```
`init_ui` allows you to declare new components from code when starting streamsync.

>>> with ss.init_ui() as ui:
>>>   with ui.Page('hello'):
>>>     ui.Text("Hello pigeons")
```

src/streamsync/__init__.py
```diff
@@ -38,3 +39,24 @@ def init_state(state_dict: Dict[str, Any]):
...
+def init_ui() -> StreamsyncUIManager:
+    """
+    Initializes and returns an instance of StreamsyncUIManager. This manager provides 
+    methods to dynamically create and manage UI components in a Streamsync application.
+
+    The returned StreamsyncUIManager allows for the creation of application-wide, 
+    code-managed components during startup, ensuring that the set of components 
+    is initially accessible by all sessions.
+
+    Returns:
```
FabienArcellier, line 53: You are using several documentation format (Google here and rEST in the other function). We should stick to one. I usually use the rEST one (https://stackoverflow.com/a/24385103) because I stick on this one usually because I am using [sphinx](https://www.sphinx-doc.org/en/master/) to write and generate API documentation.

Do you have one you prefer ?

mmikita95: My apologies, I intended to use the rEST style, but forgot to rewrite it here. Will fix it.

PR #302, opened 2024-03-12
feat: added reusable component

raaymax:
I'm not entirely satisfied with how the selection of this component works, but on the other hand, it operates similarly to the repeater component. Additionally, I'm unsure if there's any sense in using dynamic options for the parent component ID in this case. It might be easier to just manually copy and paste the component IDs.


ramedina86:
That's a good point about ids... maybe just a "copy" icon next to the id for convenience? What do you think?

raaymax:
https://github.com/streamsync-cloud/streamsync/assets/649972/733611ea-59d8-4810-814c-1477bb08fc8d

ui/src/builder/BuilderCopyText.vue
```diff
@@ -0,0 +1,69 @@
...
+
+.tooltip .tooltiptext::after {
+	content: "";
+	position: absolute;
+	top: 100%;
+	left: 50%;
+	margin-left: -5px;
+	border-width: 5px;
+	border-style: solid;
+	border-color: #555 transparent transparent transparent;
```
ramedina86, line 55: Can we avoid hardcoding here? i.e. `var(--builderActionOngoingColor);`

I'm also seeing there's a very minor difference in colour between the two.

ui/src/builder/BuilderCopyText.vue
```diff
@@ -0,0 +1,69 @@
...
+	display: inline-block;
+}
+
+.tooltip .tooltiptext {
+	visibility: hidden;
+	background-color: var(--builderActionOngoingColor);
+	color: #fff;
+	text-align: center;
+	border-radius: 20px;
+	padding: 5px 10px;
```
ramedina86, line 36: I try quite hard to use multiples of `4px` in margins, paddings, border-radiuses, widths, etc e.g. `4px 12px`. Reason is maintaining a grid and hopefully one day migrating to a more sophisticated way of managing spacing. It's less of an issue in Builder, but still I'd prefer it.

ui/src/builder/BuilderCopyText.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+	<span class="tooltip">
```
ramedina86, line 2: If we can center it on long ids that'd be good, but not a deal breaker at all.

Screenshot 2024-03-12 at 16 11 23

ui/src/builder/BuilderCopyText.vue
```diff
@@ -0,0 +1,69 @@
...
+<style scoped>
+.tooltip {
+	position: relative;
+	display: inline-block;
+}
+
+.tooltip .tooltiptext {
+	visibility: hidden;
+	background-color: var(--builderActionOngoingColor);
+	color: #fff;
```
ramedina86, line 33: Please use `--builderBackgroundColor`, looks better and if one day we get dark mode or do a Barbie edition we're set.

ui/src/core_components/other/CoreReusable.vue
```diff
@@ -0,0 +1,54 @@
...
+function render() {
+	if (!fields.parentId.value) {
+		vnode.value = h("div", "No component selected");
+		return;
+	}
+	const reusedNode = renderProxiedComponent(fields.parentId.value, 0);
+	vnode.value = reusedNode;
+}
+
+watch([fields.parentId], render);
```
ramedina86, line 44: Better to use `{ immediate: true }` when setting up the watcher.

ui/src/core_components/other/CoreReusable.vue
```diff
@@ -0,0 +1,54 @@
...
+};
+</script>
+<script setup lang="ts">
+const fields = inject(injectionKeys.evaluatedFields);
+const renderProxiedComponent = inject(injectionKeys.renderProxiedComponent);
+const vnode = ref(h("div", "No component selected"));
+
+function render() {
+	if (!fields.parentId.value) {
+		vnode.value = h("div", "No component selected");
```
ramedina86, line 37: Can we make this behave like Empty _Message_ for example? Also please specify for what purpose i.e. "No component specified for reuse.".

It shouldn't show on Preview/run (use `isBeingEdited`) injectable.

ui/src/core_components/other/CoreReusable.vue
```diff
@@ -0,0 +1,54 @@
...
+		return;
+	}
+	const reusedNode = renderProxiedComponent(fields.parentId.value, 0);
```
ramedina86, line 40: Please check if the component exists and if not show "The id specified for reuse doesn't match any component." or similar. Also shouldn't show on Preview/run.

ui/src/core_components/other/CoreReusable.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+	<div class="CoreReuse">
+		<component :is="vnode" />
```
ramedina86, line 3: 🚨🚨🚨 I think the whole thing will blow up when using _Sidebar_, which is one of the main use cases.

ramedina86 (commented):
I tried this via Scalingo. Experience is great already.

Please address the comment on _Sidebar_ first, it might be a pain.

I've added some comments to improve it a bit further, plus my usual nitpicking when it comes to any ui work.

ui/src/core_components/other/CoreReusable.vue
raaymax: Yes, it exploded for sidebar.

FabienArcellier:
Great work @raaymax . I love this component. It works well and the setup experience is interesting. I would like to suggest an improvement for the edit mode. I think the selection experience would be improved if the user couldn't select inside the component. Currently it's jumping around the component tree chaotically.

Peek 2024-03-13 08-54

Would it be possible to add an invisible overlay above the component and once the component is selected, it become visible as grey overlay and display a button that moves on click the developper to the original component?

ui/src/builder/BuilderCopyText.vue
raaymax: It is centered to span containing this text, only thing I can do here is adding:
```
white-space: nowrap;
```
but not sure if this is the way?

raaymax:
@FabienArcellier I don't think selecting nested components this way would be possible.
I could try to block selection inside of reusable component entirely and add button "go to component" or something to settings panel on right. What do you think about it?

FabienArcellier:
> @FabienArcellier I don't think selecting nested components this way would be possible. I could try to block selection inside of reusable component entirely and add button "go to component" or something to settings panel on right. What do you think about it?

That would be awesome. That's what I thought I asked

PR #304, opened 2024-03-12
tests: checking if all components have images in docs

raaymax:


docs/package.json
```diff
@@ -6,11 +6,15 @@
   "scripts": {
     "dev": "vitepress dev docs",
     "build": "vitepress build docs",
-    "preview": "vitepress preview docs"
+    "preview": "vitepress preview docs",
+		"test": "jest"
```
FabienArcellier, line 10: I think indentation is not valid

PR #307, opened 2024-03-12
feat: repeater triggers exception when the last item is removed from linked state element

### LINKED ISSUE #223: Repeater triggers exception when the last item is removed from linked state element
> There’s one trivial issue I’ve noticed with frontend – CoreRepeater.vue:75 triggers an exception when we try to remove the last element from the state linked to Repeater – probably due to that it’s trying to extract keys from null.
> -- @mmikita95 

Mikita, thanks for reporting this, I'll deal with it -assigning you for visibility.

FabienArcellier:
The repeater is invisible when the application is running and the last element has been deleted. The default slot remains visible in edit mode.

fix #223 (maybe)

Peek 2024-03-12 21-58


ramedina86 (approved):
Nice one, I like the added behaviour of distinguishing between editing and preview/run.

PR #313, opened 2024-03-16
fix: added support for autocompletion on the low ui approach for pycharm

FabienArcellier:
Pycharm does not currently support autocompletion on TypedDict passed as an argument to a method. This problem does not seem to be really addressed in their roadmap.

https://youtrack.jetbrains.com/issue/PY-54151/TypedDict-completion-at-callee-does-not-work-for-methods
On the other hand, Pycharm supports autocompletion on static methods of a class.

I propose a draft to manage the component tree as a context variable. This pattern offer 2 advantages :

* all the methods of ui manager can be static
* a developper than build an extension can manipulate the component tree easily inside its extension

---
https://github.com/FabienArcellier/streamsync/issues/40

---
Peek 2024-03-16 05-27

---
I test this implementation on event handler and async event handler

### DIFF
--- a/src/streamsync/ui_manager.py
+++ b/src/streamsync/ui_manager.py
@@ -1,9 +1,8 @@
 from json import dumps as json_dumps
-from typing import Optional, Union
+from typing import Optional
 
-from streamsync.core import base_component_tree
-from streamsync.core_ui import (Component, SessionComponentTree, UIError,
-                                current_parent_container)
+from streamsync.core_ui import (Component, UIError,
+                                current_parent_container, current_component_tree, ComponentTree)
 
 
 class StreamsyncUI:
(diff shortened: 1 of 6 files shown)


src/streamsync/ui_manager.py
```diff
@@ -30,102 +24,119 @@ def assert_in_container():
         if container is None:
             raise UIError("A component can only be created inside a container")
 
+    @property
+    def component_tree(self) -> ComponentTree:
```
mmikita95, line 28: I see that this property is not used, and other methods still access `current_component_tree()` directly. I believe we could replace all calls to it with this property, or is there something I am missing?

FabienArcellier: I keep it for debug purpose on the IDE.

FabienArcellier: Finally we can't because the method are static and component tree is a property. It's not accessible

PR #317, opened 2024-03-18
chore: check if ui binding is up to date in continuous integration

FabienArcellier:
check in continuous integration if the ui binding is up to date. If it's not, it raise an error and show the following error message to user :

```
UI binding is incomplete and has to be regenerated with `alfred npm.codegen.binding.ui`
```


raaymax (approved):
Nice one! 🎉

PR #319, opened 2024-03-18
chore: check if documentation is building in the continuous integration on pipeline

FabienArcellier:
The streamsync site build and documentation is not verified in continuous integration. I noticed that the documentation in dev currently does not build correctly.

* [x] add the command `alfred docs.build`
* [x] run the command `alfred docs.build` in `alfred ci`

image

### DIFF
--- a/alfred/ci.py
+++ b/alfred/ci.py
@@ -11,16 +11,24 @@ import alfred
 @alfred.option('--front', '-f', help="run for frontend only", is_flag=True, default=False)
 @alfred.option('--back', '-b', help="run for backend only", is_flag=True, default=False)
 @alfred.option('--e2e', '-e', help="run for end-to-end only", default=None)
-def ci(front, back, e2e):
-    if back or (not front and not back and not e2e):
+@alfred.option('--docs', '-e', help="run for documentation only", default=False)
+def ci(front, back, e2e, docs):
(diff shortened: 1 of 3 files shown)


ramedina86:
Won't this take ages? Will it even run in one of the Github Actions containers?

@raaymax will probably know better than I do though

FabienArcellier:
> Won't this take ages? Will it even run in one of the Github Actions containers?
> 
> @raaymax will probably know better than I do though

The step is taking 5 seconds. It's already running in this PR.

image

raaymax:
I think docs build where also added in this one https://github.com/streamsync-cloud/streamsync/pull/286 but in slightly different way it was added to `npm run build`.

FabienArcellier:
> I think docs build where also added in this one #286 but in slightly different way it was added to `npm run build`.

I am hesitant. Documentation has it's own scope so it's logic to have dedicated build step. In same time, having everything run when executing `npm run build` can make sense. You don't have to think about.

raaymax:
My proposition would be to have `npm run build` to build everything and in alfred lets use separate build steps for each package? i already updated https://github.com/streamsync-cloud/streamsync/pull/286 to reflect this.

FabienArcellier:
Ok, so I close this PR

PR #320, opened 2024-03-18
docs: code managed components

mmikita95:
Documentation describing the usage of features associated with code-managed components.


docs/docs/modifying-app-ui-through-backend.md
```diff
@@ -0,0 +1,215 @@
...
+    ```
+    *Position is calculated automatically for each component, and you should be careful when you override it with predefined value, as this might lead to unexpected results.*
+- **`parentId: str`**: Determines the parent [container](#container-components) for the component. By default, components recognize the container in the context of which they were defined as their parent. This allows for linking components to their parents outside of context, or for overriding a parent within a context.
+    ```python
+    ui.Text(
+        {"text": "Hello Parent, I'm your child too!"}, 
+        parentId="dear-parent"
+        )
+    ```
+- **`visible: bool`**: Determines the visibility of the component, `True` by default.
```
ramedina86, line 127: Visible should be `bool | str` / `Union[bool, str]`. Hope that's a doc oversight!

ramedina86 (commented):
💯

docs/docs/modifying-app-ui-through-backend.md
mmikita95: Yeah, my bad 😅

PR #328, opened 2024-03-19
chore: preset IDs for pages

mmikita95:
Adding a mechanic to keep the count of pages in component trees, and to add predefined keys & IDs for pages if case if they're not provided by user. Fixes a bug with losing CMC page focus upon app code update.

### DIFF
--- a/src/streamsync/core_ui.py
+++ b/src/streamsync/core_ui.py
@@ -44,6 +44,7 @@ class ComponentTree:
 
     def __init__(self, attach_root=True) -> None:
         self.counter: int = 0
+        self.page_counter: int = 0
         self.components: Dict[str, Component] = {}
         if attach_root:
             root_component = Component(
@@ -83,6 +84,8 @@ class ComponentTree:
 
     def attach(self, component: Component, override=False) -> None:
         self.counter += 1
+        if component.type == "page":
+            self.page_counter += 1
         if (component.id in self.components) and (override is False):
             raise RuntimeWarning(f"Component with ID {component.id} already exists")
         self.components[component.id] = component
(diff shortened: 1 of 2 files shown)


src/streamsync/core_ui.py
```diff
@@ -100,6 +103,8 @@ def ingest(self, serialised_components: Dict[str, Any]) -> None:
             self.components.pop(component_id)
         for component_id, sc in serialised_components.items():
             component = Component(**sc)
+            if component.type == "page" and component_id not in self.components:
```
FabienArcellier, line 106: Did you think to reuse attach here ?

src/streamsync/ui_manager.py
```diff
@@ -137,6 +136,13 @@ def _create_component(component_tree: ComponentTree,  component_type: str, **kwa
     raw_content: dict = kwargs.pop("content", {})
     content = {key: _prepare_value(value) for key, value in raw_content.items()}
 
+    if component_type == "page" and "id" not in kwargs:
```
FabienArcellier, line 139: Add a comment block that describe why we set automatic id on Page and not on other component

FabienArcellier (approved):
I have two remarks

src/streamsync/core_ui.py
mmikita95: This was what I've tried initially, but it leads to constant re-attachment of the whole tree, including root components. Paints console red, to put it shortly 🥲

ramedina86:
@FabienArcellier @mmikita95 are we good to merge?

mmikita95:
@ramedina86 yes, good to go.

PR #331, opened 2024-03-19
refactor: Change flag to isCodeManaged

ramedina86:
I refactored `flag` to `isCodeManaged` to improve readability.

Also,

**Frontend:**

Some position-related fixes:

- Insertion position in `useComponentActions` is independent of CMCs
- Higher siblings in `useComponentActions` don't have their positions incremented. Before, when inserting a BMC before a CMC, that'd increment the position of the CMC.
- AuditAndFix will ignore CMCs when fixing positioning.
- I added a more descriptive `getComponentChildren` to get children, with filters for BMC and CMC, also orders BMC first and CMC later. I intend to replace all references of `getComponents(parentId)` after release.

**Backend:**

- I made a change in the backend's `determine_position` to ignore BMCs. We now have, for example, two components with position 0, but BMCs are always rendered first.
- Removed unused `get_direct_descendents_length`

### DIFF
--- a/ui/src/builder/BuilderSettings.vue
+++ b/ui/src/builder/BuilderSettings.vue
@@ -74,7 +74,7 @@ const ssbm = inject(injectionKeys.builderManager);
 const docsActive = ref(false);
 
 const component = computed(() => ss.getComponentById(ssbm.getSelectedId()));
-const readonly = computed(() => component.value.flag === "cmc");
+const readonly = computed(() => component.value.isCodeManaged);
 
 const componentDefinition = computed(() => {
 	const { type } = component.value;
--- a/ui/src/core/index.ts
+++ b/ui/src/core/index.ts
@@ -510,6 +510,8 @@ export function generateCore() {
 	 * Gets children of a specific Streamsync component.
 	 *
 	 * @param parentId The parent id.
+	 * @param includeBMC Whether to include Builder-managed components
+	 * @param includeCMC Whether to include code-managed components
 	 * @returns An array of components.
 	 */
 	function getComponentChildren(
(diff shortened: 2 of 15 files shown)


ui/src/core/index.ts
```diff
@@ -506,6 +506,36 @@ export function generateCore() {
...
+	 * @param parentId The parent id.
+	 * @param includeBMC Whether to include Builder-managed components
+	 * @param includeCMC Whether to include code-managed components
+	 * @returns An array of components.
+	 */
+	function getComponentChildren(
+		parentId: Component["id"],
+		includeBMC: boolean = true,
+		includeCMC: boolean = true,
+	): Component[] {
```
raaymax, line 521: suggestion: I think having `getComponents` and `getComponentChildren` will be confusing. Maybe it would be a good idea to merge those 2 functions into one? 
```suggestion
	function getComponents(
		parentId: Component["id"],
		{
			includeBMC = true,
			includeCMC = true,
			sortedByPosition = false,
		}: {
			includeBMC?: boolean;
			includeCMC?: boolean;
			sortedByPosition?: boolean;
		} = {},
	): Component[] {
```

ui/src/builder/BuilderSettings.vue
```diff
@@ -74,7 +74,7 @@ const ssbm = inject(injectionKeys.builderManager);
 const docsActive = ref(false);
 
 const component = computed(() => ss.getComponentById(ssbm.getSelectedId()));
-const readonly = computed(() => component.value.flag === "cmc");
+const readonly = computed(() => component.value.isCodeManaged);
```
raaymax, line 77: question: is computed here still required? 
Can't we use this `v-if="component.isCodeManaged"`

raaymax (commented):
This change is definitely highly desirable! Nice one!

ui/src/core/index.ts
ramedina86: I liked your "options" object argument. I'll go with that.

My idea was to leave `getComponents` when we need to get all components, and `getComponentChildren` when we need the children of a component.

I'll go for the proposed syntax and unify under `getComponents`. `parentId` needs to be optional though, as sometimes we get everything.

ui/src/builder/BuilderSettings.vue
ramedina86: I left it because it helps with readability, but probably overkill.

PR #336, opened 2024-03-21
feat: check CMC parent-child relations

mmikita95:
Depends on #335 

Verifies whether a CMC can be a valid child for provided `parent_id`, and whether a parent component associated with this id can be a valid parent for the CMC.

### DIFF
--- a/src/streamsync/ui_manager.py
+++ b/src/streamsync/ui_manager.py
@@ -137,13 +137,17 @@ def _check_parent_child_relations(
             )
 
     valid_children_types_for_parent = \
-        StreamsyncUIManager.children_types_map.get(parent.type)
+        StreamsyncUIManager.children_types_map.get(parent.type, [])
 
     while "inherit" in valid_children_types_for_parent:
         # Switch to grandparent allowed types in case of "inherit" instruction
+        if not parent.parentId:
+            break
         grandparent = component_tree.get_component(parent.parentId)
+        if not grandparent:
+            break
         valid_children_types_for_parent = \
-            StreamsyncUIManager.children_types_map.get(grandparent.type)
+            StreamsyncUIManager.children_types_map.get(grandparent.type, [])
 
     valid_parent_types_for_component = \
         StreamsyncUIManager.parent_types_map.get(component_type)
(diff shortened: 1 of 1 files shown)


src/streamsync/ui_manager.py
```diff
@@ -115,9 +117,57 @@ def _prepare_value(value):
...
+        parent_id: str,
+        component_type: str
+):
+    # Import required inside a function:
+    # StreamsyncUIManager class stores actual type maps,
+    # but importing it directly causes a circular import
+    from streamsync.ui import StreamsyncUIManager
+
+    if component_type == "root":
+        raise UIError("Root component cannot be a child component.")
```
mmikita95, line 131: Root check is required to avoid the case when 'root' type can "slip" through a wildcard (`*`).

src/streamsync/ui_manager.py
```diff
@@ -115,9 +117,57 @@ def _prepare_value(value):
...
+       or not valid_parent_types_for_component:
+        type_to_blame = (f"Parent type '{parent.type}'"
+                         if not valid_children_types_for_parent
+                         else f"Child type '{component_type}'")
+        raise RuntimeError(
+            f"Misconfigured types: {type_to_blame} is not present " +
+            "in allowed types map."
+            )
+
+    is_component_eligible = \
```
ramedina86, line 154: Are we checking "inherit"?

mmikita95: No, my apologies, I wasn't aware of it. Can you explain it a bit so I can account for it?

mmikita95: Ah, I think I understand now. If we're talking `Repeater`, we need to check on its parent for valid types.

ramedina86: You have the frontend logic here https://github.com/streamsync-cloud/streamsync/blob/dev/ui/src/core/typeHierarchy.ts

For example, if you have a _Column Container_, which inside has a _Repeater_, which in turn has another _Repeater_ inside, the 2nd (nested) _Repeater_ should only allow _Column_, _Repeater_ and I believe _Separator_.

It makes things harder. I'm not sure if it's worth replicating this in the backend. I'd do it in the frontend, which would also helps us with `ui.json` manipulation/corruption.

mmikita95:
As we're going to do this kind of validation on frontend, closed

PR #337, opened 2024-03-21
feat: added rendering error when component can't be rendered

raaymax:


ui/src/renderer/ComponentProxy.vue
```diff
@@ -310,11 +311,25 @@ export default {
 			const vnodeProps = {
 				...getRootElProps(),
 			};
-			const renderedComponent = h(template, vnodeProps, {
-				default: defaultSlotFn,
-			});
 
-			return renderedComponent;
+			const parentId = instancePath.at(-2)?.componentId;
```
ramedina86, line 315: Nice to see you decoded `instancePath`, but no need to flex, I think we can just use `parentId` :)

ui/src/renderer/ComponentProxy.vue
```diff
@@ -310,11 +311,25 @@ export default {
...
+			const allowedTypes = !parentId
+				? ["root"]
```
ramedina86, line 317: Logic looks good, but not the most readable, can you refactor a bit? 

`if (isParentSuitable()) {
  return render normally...
}

return getRenderErrorVNode(type, "parent not suitable");
`

ui/src/renderer/ComponentProxy.vue
```diff
@@ -310,11 +311,25 @@ export default {
...
+				: ss.getContainableTypes(parentId);
+			if (allowedTypes.includes(component.value.type)) {
+				return h(template, vnodeProps, {
+					default: defaultSlotFn,
+				});
+			} else {
+				if (!isBeingEdited.value) return h("div");
+				return [
```
ramedina86, line 325: I think the `[]` you can just pass the `h` and save a level of indentation.

ui/src/renderer/ComponentProxy.vue
```diff
@@ -310,11 +311,25 @@ export default {
...
+			const parentId = instancePath.at(-2)?.componentId;
+			const allowedTypes = !parentId
+				? ["root"]
+				: ss.getContainableTypes(parentId);
+			if (allowedTypes.includes(component.value.type)) {
+				return h(template, vnodeProps, {
+					default: defaultSlotFn,
+				});
+			} else {
+				if (!isBeingEdited.value) return h("div");
```
ramedina86, line 324: I'd do a `getRenderVNode` function as mentioned in the other component and as part of that do the

`if (!isBeingEdited.value) return h("div");`

Since it's probably something we want to do for every error type? (I know we only have this one, but I liked the `RenderError` idea)

PR #350, opened 2024-03-25
generate a documentation page for every components

FabienArcellier:
Generate a dedicated space to navigate inside existing components

Peek 2024-03-25 18-30


docs/docs/components/avatar.md
```diff
@@ -0,0 +1,77 @@
+<!-- This file is generated through a template. Do not edit directly. -->
+<!-- generated by ui/tools/generator_doc_components_page.js  -->
+<script setup>
+	import { data as components } from "../components.data.ts";
+	const component = components.find(c => c.name === "Avatar");
+</script>
+
+<h1>{{ component.name }}</h1>
```
FabienArcellier, line 8: I should use the markdown inclusion to keep the generated part as light as possible

https://vitepress.dev/guide/markdown#markdown-file-inclusion

ui/components.json
FabienArcellier, line 1: rename this file into `components.codegen.json` as discussed this morning

docs/docs/core.ts
raaymax, line 1: I think these comments don't bring too much value.

ui/tools/generator_ui_component_json.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @returns {Promise<void>}
+ */
+async function generate() {
+	const components = await loadComponents();
+
+	components.forEach((component) => {
+		// eslint-disable-next-line prettier/prettier
+		const componentFile = 'Core' + component.type[0].toUpperCase() + component.type.slice(1) + '.vue';
+		component.source_link = `ui/src/core_components/${component.category.toLowerCase()}/${componentFile}`;
```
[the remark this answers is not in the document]
raaymax: Only other solution I can come up with is adding something like this
```javascript
		_source: import.meta.url.split("core_components")[1],
```
to every component we have

ui/tools/generator.js
```diff
@@ -1,160 +1,12 @@
...
-            position=position,
-            parentId=parentId,
-            handlers=handlers,
-            visible=visible${isBindable ? bindPass : ""})
-        return component
-    `;
-	});
-	return methods.join("");
+if (require.main === module) {
+	generate();
```
raaymax, line 11: The if statement here is unnecessary. Just call the function.

FabienArcellier: I just reproduce a python pattern per habits. It allow to import any module as library without running the code inside.

FabienArcellier:
I have to edit the option with `yes` and `no`.

image

FabienArcellier:
The options should be visible in the comment after attributes

image

ui/tools/generator_ui_component_json.js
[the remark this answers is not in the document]
FabienArcellier: Thanks for the feedback. I think I'll keep this code and add automatic tests to enforce the component file tree structure. I am convinced adding this line into all the component.

ui/tools/generator.js
raaymax: I know but in this file you don't have any exports so there is not point importing this file as a library.

FabienArcellier:
add the documentation field when it's present below the image

docs/docs/core.ts
```diff
@@ -0,0 +1,143 @@
...
+	});
+
+	return categoriesAll;
+}
+
+/**
+ * Get the description of a category
+ * @param name Name of the category
+ */
+export function categoryDescription(name: string) {
```
raaymax, line 35: I mean comments like this, whats the purpose of them? 
This is how I see them:
Screenshot 2024-03-27 at 11 53 34
😅 
But maybe there is some hidden reason for them?

ui/tools/generator_ui_component_json.js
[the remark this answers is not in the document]
raaymax: I think I found even better solution! I need to test it first tho...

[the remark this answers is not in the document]
ramedina86: Great job with this @raaymax , the original solution would have indeed been rather fragile.

ramedina86:
> I have to edit the option with `yes` and `no`.
> 
> image

What did you mean by this?

ramedina86:
@FabienArcellier can you please merge dev into your branch? Latest changes by Mateusz create conflicts here:

```
ui/components.json
ui/src/core_components/content/CoreMessage.vue
ui/tools/generator.js
```

FabienArcellier:
> > I have to edit the option with `yes` and `no`.
> > image
> 
> What did you mean by this?

It was not practical to have different order when reading it.

PR #353, opened 2024-03-27
refactor: separate trees

mmikita95:
- introducing a `DependentComponentTree` class to generalize relations between base tree and base CMC tree/session tree
- introducing a `base_cmc_tree` to accept CMCs created via `ss.init_ui()`
- allows for removing components from dependent trees, but raises a `UIError` if trying to remove BMCs
- dependent tree gathers direct descendants both on base tree and its own set of components – fixes positioning bug
- filtering out CMCs during ingestion is not required anymore, as frontend does not send them to backend

Functions well as of now, but further updates would require separation of stored components on frontend as well – this will allow us to avoid using the flag.


src/streamsync/core_ui.py
```diff
@@ -157,11 +182,41 @@ def to_dict(self) -> Dict:
...
+    def __init__(self, base_component_tree: ComponentTree, base_cmc_tree: ComponentTree):
+        super().__init__(base_component_tree, attach_root=False)
+
+        # Initialize SessionComponentTree with components from the base
+        # CMC pool, added during app initialization
+        preinitialized_components = base_cmc_tree.to_dict()
+        self.ingest(preinitialized_components)
+        self.counter = len(self.components)
+
+        preinitialized_pages = \
```
ramedina86, line 202: Do we need this?

If we have two BMC pages, I'm happy for the first CMC page to be named # cmc-page1 (or similar)

So I'm okay with initialising counter to 0

src/streamsync/core_ui.py
```diff
@@ -157,11 +182,41 @@ def to_dict(self) -> Dict:
...
+class SessionComponentTree(DependentComponentTree):
+
+    def __init__(self, base_component_tree: ComponentTree, base_cmc_tree: ComponentTree):
+        super().__init__(base_component_tree, attach_root=False)
+
+        # Initialize SessionComponentTree with components from the base
+        # CMC pool, added during app initialization
+        preinitialized_components = base_cmc_tree.to_dict()
+        self.ingest(preinitialized_components)
+        self.counter = len(self.components)
```
ramedina86, line 200: What are we using the counter for? I remember I set it up initially, but I'm not sure if it's used

src/streamsync/core_ui.py
mmikita95: We probably need some good definitions for these concepts 😅
This `base_cmc_tree` actually contains CMCs, including CMC pages, that were created with `ss.init_ui()`. So yes, we need to count them and to carry this counter to the session tree, so we can safely create pages from handlers and all IDs end up unique (whereas if this was 0, we could have cmc-page-1 made via `ss.init_ui()` and cmc-page-1 made via handler.

src/streamsync/core_ui.py
mmikita95: To be honest, I haven't seen it being used, but kept it just to be safe in case I missed it somewhere.

src/streamsync/core_ui.py
ramedina86: Great point, thanks for the explanation.

src/streamsync/core_ui.py
ramedina86: Cool, can you please do a `.counter` search and see how that goes? That should cover it

mmikita95: image
Seems to be confined to these classes indeed, safe to remove.

ramedina86: Nice, let's remove

PR #364, opened 2024-03-28
docs: added docs custom tag to vue sfc and file path for components.json

raaymax:


ui/vite.config.js
```diff
@@ -4,10 +4,44 @@ import { defineConfig } from "vite";
 import vue from "@vitejs/plugin-vue";
 import monacoEditorPlugin from "vite-plugin-monaco-editor";
 
+const streamsyncComponentsDocs = () => {
```
ramedina86, line 7: The idea is amazing and the templates look so much cleaner.

This config looks a bit dirty now though. Can you please refactor by:
- Taking this to another file and importing
- Make it obvious that it's a plugin 

(similarly to `monacoEditorPlugin`)

ui/vite.config.js
```diff
@@ -4,10 +4,44 @@ import { defineConfig } from "vite";
 import vue from "@vitejs/plugin-vue";
 import monacoEditorPlugin from "vite-plugin-monaco-editor";
```
ramedina86, line 6: Can you please look into having the Vite config as TypeScript? Pretty sure a `vite.config.ts` is possible.

I'm not sure why I created a JS config.

ui/vite.config.js
```diff
@@ -4,10 +4,44 @@ import { defineConfig } from "vite";
...
+			}
+		},
+	};
+};
+
 // https://vitejs.dev/config/
 export default defineConfig({
 	base: "./",
-	plugins: [vue(), monacoEditorPlugin({})],
+	plugins: [vue(), monacoEditorPlugin({}), streamsyncComponentsDocs()],
```
ramedina86, line 43: Does this work for custom components? The plugin probably needs to be added there too.

ramedina86: `vite.config.custom.js` that is.

ui/vite.config.js
```diff
@@ -4,10 +4,44 @@ import { defineConfig } from "vite";
 import vue from "@vitejs/plugin-vue";
 import monacoEditorPlugin from "vite-plugin-monaco-editor";
 
+const streamsyncComponentsDocs = () => {
+	let config;
+
+	return {
+		name: "vue-docs-ignore",
```
ramedina86, line 11: Is `vue-docs-ignore` descriptive? Aren't we transforming rather than ignoring? Please correct me if I'm wrong.

ramedina86 (approved):
Looking great

Issue #291, opened 2024-03-11
add a tooltip on icon, button, ... component


FabienArcellier:
Components that have an `ss-click` event should have a setting to configure a tooltip on mouseover.

![image](https://github.com/streamsync-cloud/streamsync/assets/159559/bdb24d46-d044-435d-a2b2-00a67c8a71b1)

If the tooltip field is empty, nothing appears. The components work as they do today.

#### Components

* Button
* Text
* Image
* Icon (even if it doesn't have the `ss-click`)

#### Design proposal

A simple tooltip above the button as in material ui.

![Peek 2024-03-11 11-31](https://github.com/streamsync-cloud/streamsync/assets/159559/46af5294-d4be-4baa-988e-64696823dfa4)

PR #366, opened 2024-04-01
feat: add tooltip

oandalib:
This commit adds tooltip feature per #291.

https://github.com/streamsync-cloud/streamsync/assets/24489388/34cd6850-e24a-420d-8f9a-cb4387310bc2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1288,6 +1288,27 @@
 				"node": "^12.22.0 || ^14.17.0 || >=16.0.0"
 			}
 		},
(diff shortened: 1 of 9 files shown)


FabienArcellier:
@oandalib could you run npm install to generate the `package-lock.json` ? You add a library but it has not been locked as a dependency. It should solve the current error message you have in the continuous integration.

FabienArcellier:
We take time to merge this one, because we have discussion about adding `floating-vue` as an external dependency. We currently try to avoid external css dependencies when we can.

PR #367, opened 2024-04-01
refact: proposal for a new design arround component tree

FabienArcellier:
This is a refactoring proposal to make the notion of component tree simpler. This refactoring removes the 3 distinct classes that inherit from each other to represent the 3 trees. It removes cascading inheritance and method overloading.

There is a single class `ComponentTree` class that can combine fragments. One fragment for BMC, one for CMC, one for SMC. A fragment can be frozen to lock its contents. Fragments are an instance of the same `ComponentTreeFragment` class.

The application contains 2 tree construction methods. `build_base_component_tree` which builds the tree the application needs upon loading with a fragment for BMCs and CMCs. `build_session_component_tree` builds a tree dedicated to the user session with 3 fragments, one fragment for SMCs, one for CMCs and a frozen fragment for BMCs. The fragments are copied to separate them from the base representation.

I added in the tests, 2 distinct contexts, one to represent a ui manager when streamsync application is loading and one to represent a ui manager in an event handler.

### DIFF
--- a/src/streamsync/core_ui.py
+++ b/src/streamsync/core_ui.py
@@ -78,6 +78,21 @@ class ComponentTreeBranch:
         self.component_tree_id = id
         self.freeze = freeze
 
+    def clone(self) -> 'ComponentTreeBranch':
+        """
+        Clone a component branch into a new one.
+
+        Streamsync uses this action when it instantiates the component tree attached to the session.
+        This ensures complete insulation of the original shaft.
+
+        >>> cloned = bmc_tree.clone()
+        :return:
+        """
+        cloned = ComponentTreeBranch(self.component_tree_id, self.freeze)
+        cloned.components = copy.copy(self.components)
+        cloned.page_counter = self.page_counter
+        return cloned
+
 
     def get_component(self, component_id: str) -> Optional[Component]:
         return self.components.get(component_id)
@@ -341,8 +356,8 @@ def build_session_component_tree(base_component_tree: ComponentTree) -> Componen
     """
     session_tree_branch = ComponentTreeBranch(Branch.session_cmc)
 
-    cmc_tree_branch = copy.copy(base_component_tree.branch(Branch.initial_cmc))
-    bmc_tree_branch = copy.copy(base_component_tree.branch(Branch.bmc))
+    cmc_tree_branch = base_component_tree.branch(Branch.initial_cmc).clone()
+    bmc_tree_branch = base_component_tree.branch(Branch.bmc).clone()
     bmc_tree_branch.freeze = True
 
     return ComponentTree([session_tree_branch, cmc_tree_branch, bmc_tree_branch])
--- a/src/streamsync/core_ui.py
+++ b/src/streamsync/core_ui.py
@@ -25,14 +25,14 @@ class Branch(Enum):
     Enum for the component tree branches that can be created in streamsync
 
     * bmc: builder managed component
-    * cmc: code managed component
+    * initial_cmc: code managed component
     * session: session managed component
 
     This enum should be used only in the module core_ui.py.
     """
     bmc = "bmc"
-    cmc = "cmc"
-    session = "session"
+    initial_cmc = "initial_cmc"
+    session_cmc = "session_cmc"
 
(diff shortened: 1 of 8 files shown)


src/streamsync/core_ui.py
```diff
@@ -116,117 +118,244 @@ def to_dict(self) -> Dict:
...
 
 
-class DependentComponentTree(ComponentTree):
+class ComponentTree():
 
-    def __init__(self, base_component_tree: ComponentTree, attach_root=False):
-        super().__init__(attach_root)
-        self.base_component_tree = base_component_tree
+    def __init__(self, trees: List[ComponentTreeFragment]):
+        assert len(trees) > 0, "Component tree must have at least one tree fragment"
```
mmikita95, line 128: Not necessarily true for CMC trees, they can be empty if the user chooses to not create any CMCs

src/streamsync/core_ui.py
```diff
@@ -116,117 +118,244 @@ def to_dict(self) -> Dict:
...
+        self.trees = trees
```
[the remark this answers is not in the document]
FabienArcellier: Yes that would make sense

src/streamsync/core_ui.py
```diff
@@ -116,117 +118,244 @@ def to_dict(self) -> Dict:
...
+    session_fragment = ComponentTreeFragment(Fragment.session)
+
+    cmc_tree_fragment = copy.copy(base_component_tree.fragment(Fragment.cmc))
+    bmc_tree_fragment = copy.copy(base_component_tree.fragment(Fragment.bmc))
+    bmc_tree_fragment.freeze = True
+
+    return ComponentTree([session_fragment, cmc_tree_fragment, bmc_tree_fragment])
+
+
+def injest_bmc_component_tree(component_tree: ComponentTree, components: Dict[str, Any]):
```
mmikita95, line 320: I believe there's a typo, "injest => ingest"?

src/streamsync/core_ui.py
FabienArcellier: I renamed trees to tree_branches. A tree can have 0 components but it must still have several branches, one for the bmc, one for the cmc and one for the smc

ramedina86:
This is a step in the right direction. However, I'd like to change the nomenclature "smc" as I don't think it's reflective of what we're doing.

**Manage** should be interpreted in the sense of enabling the user to create, delete, etc. Not from an internal perspective. Two categories emerge: 
- Builder-managed
- Code-managed

The other classification emerges from the specificity of the components:
- Shared/base
- Session-specific

Therefore, the combinations are:
- Builder-managed base components
- Code-managed base components
- Builder-managed session-specific components
- Code-managed session-specific components

There are currently no plans to have builder-managed session-specific components. For simplicity, we can adopt the following:
- Builder-managed base components: `bmc`
- Code-managed base components: `cmc`
- Code-managed session-specific components: `session-cmc`

@mmikita95 fyi, feel free to add to the conversation

FabienArcellier:
> This is a step in the right direction. However, I'd like to change the nomenclature "smc" as I don't think it's reflective of what we're doing.

That's sound interesting, I will take a look at the change involved.

mmikita95:
@ramedina86 great breakdown, thank you! 

One thing I would probably adjust is to shift focus on "session CMCs" as "main CMCs" (i.e., use names like "pre-session CMCs vs session CMCs or in-session CMCs"), because in my opinion, while "pre-session CMCs" are expected to be used more widely, technically the session context is the last point that collects all components into a "render package", and one can view "pre-session CMCs" as "copies made for the session". Therefore, to better reflect this technicality, I'd suggest this naming convention. Sorry if this is turning out to be too confusing 😅

ramedina86:
Excellent point @mmikita95 , glad I tagged you. With this in mind, what classification (short names) would you go for?

ramedina86:
`iscmc` sounds too much like a boolean. The `pre` is slightly confusing to me. How about:

- `bmc`
- `initial_cmc`
- `session_cmc`

in line with state (`initial_state`, `session_state`).

PR #368, opened 2024-04-01
feat: capture the component that trigger an event in context

FabienArcellier:
On a click or an event, a developer can consult the identifier at the origin of the event from the context. This allows it to adapt the behavior of its event handler or to retrieve the parent to add interface components.

```python
def increment(state, context):
    print(context['target'])
```

```python
def increment(state, ui, context):
   container = ui.parent(context['target'])
    with ui.find(container):
      pass
```

```python
def increment(state, ui, context):
   container = ui.parent(context['target'], level=3) # third parent
    with ui.find(container):
      pass
```

### definition of done

* add a documentation section about `Context` into [EventHandler](https://www.streamsync.cloud/event-handlers.html)

### DIFF
--- a/src/streamsync/core.py
+++ b/src/streamsync/core.py
@@ -988,6 +988,9 @@ class Evaluator:
             context[key_variable] = repeater_items[instance_number][0]
             context[value_variable] = repeater_items[instance_number][1]
 
+        if len(instance_path) > 0:
+            context['target'] = instance_path[-1]['componentId']
+            
         return context
 
     def set_state(self, expr: str, instance_path: InstancePath, value: Any) -> None:
--- a/docs/docs/backend-driven-ui.md
+++ b/docs/docs/backend-driven-ui.md
@@ -62,6 +62,18 @@ with ui.find("column-container"):
 ```
 If the component couldn't be found, the method raises a `RuntimeError`.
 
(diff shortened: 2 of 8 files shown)


src/streamsync/core.py
```diff
@@ -963,6 +963,9 @@ def get_context_data(self, instance_path: InstancePath) -> Dict[str, Any]:
             context[key_variable] = repeater_items[instance_number][0]
             context[value_variable] = repeater_items[instance_number][1]
 
+        if len(instance_path) > 0:
```
FabienArcellier, line 966: I should add a type definition (TypedDict) on context to allow autocompletion in the IDE

ramedina86:
Really nice feature.

Only thing that I'd change is the quotes, would love to see consistency in the use of double quotes, across docs and code. In terms of code. Where are we in terms of enforcing these via linter? I believe @raaymax already set up Ruff. Is it a matter of setting it up?

FabienArcellier:
> I believe @raaymax already set up Ruff. Is it a matter of setting it up?

Yes, I think we have to set it up : https://docs.astral.sh/ruff/settings/#lint_flake8-quotes_inline-quotes. I have a PR to do for ruff. I will handle this config.

PR #372, opened 2024-04-02
feat: register external modules to use its functions as event handlers

mmikita95:
This is probably subject to major changes, so adding this as a draft for now, to discuss and share ideas.

In streamsyncuserapp module, this adds an `__internal_handler_registry__`, which is a dictionary of handler maps for each registered module, including `main` that represents the `main.py` of the user's app.

Adds a `ss.init_handlers()` method, which accepts a module or a list of modules, the functions of which are registered into the corresponding handler map. 

For now, frontend receives the full notation for each handler, which is then used to retrieve the handler from the map.


ramedina86:
- I don't see a reason to hack the user module. Can't we just save things somewhere else (probably in the core) without having to hide anything?

- Can we consider using `__qualname__`?

https://levelup.gitconnected.com/what-is-qualname-in-python-and-when-to-use-it-21e84aa47a45#:~:text=In%20Python%2C%20the%20__qualname,identifiable%20in%20the%20Python%20namespace.

- I don't like the ":" as a separator and I'd rather use the qualified name

mmikita95:
> I don't see a reason to hack the user module. Can't we just save things somewhere else (probably in the core) without having to hide anything?
My logic was that "streamsyncuserapp" already does (in a way) "register" event handlers, and it is globally accessible without the risk of circular imports (which using core may present) – so I tied this mechanism to it. And, of course, this is "behind the stages", that's why I made it "hidden". But I see your point, and I'll try to modify it such.
>Can we consider using __qualname__?
Absolutely! Will do.

FabienArcellier:
Thank you @mmikita95 to share this draft. I took some time to think about it because the use of the `streamsyncuserapp` module seemed inappropriate to me.

Here is a complete answer that maybe goes beyond simple code review. This comment is long because I need to come back to the structure of Streamsync and in particular the place of the AppProcess. The changes ares minor. I describe them in the second part: **AppProcess is the kernel of streamsync application**.

I let you decide if some insight make sense or if the original proposal should be kept as is :)

#### TL;DR

Here is the change I would recommand

* the list of event handlers be a property / subproperty of the app process
* the app process should be exposed as internal api as ``get_app_process``
*  `streamsyncuserapp` module is not used anywhere except in code loading

#### Regards to Event handler management

I would recommend that **the list of events be a property / subproperty of the app process**. The use of `streamsyncuserapp` is a necessity of Python because it requires a module name to load code but must not contaminate the rest of the code. 

The use of `streamsyncuserapp` in `_get_user_functions` is a shortcut due to the way event handlers are loaded today. However, by introducing `streamsync.init_handlers`, the logic and philosophy completely changes. You can inject event handlers knowledget into AppProcess instead to relying on future module inspection.

Here is a draft of what I would expect to find when reading the code.

```python
class EventHandler(TypedDict):
    handler: Callable
    name: str
    args: List[str]

class AppProcess(multiprocessing.Process):
    # ...
    # could also be encapsulated in a sub object that handle things related to code. I would prefer generally to have dozen of properties than complex nested structure with 1-1 relation. That's a point of discussion
    self.event_handlers: Dict[str, 'EventHandler'] = {} 
    self.run_code: str = ""
    # ...

    def _get_user_functions(self) -> List[Dict]:
       return list(self.event_handlers.values())

    def _execute_user_code(self) -> None:
        # ...
        with redirect_stdout(io.StringIO()) as f:
            code = compile(self.run_code, code_path, "exec")
            exec(code, streamsyncuserapp.__dict__)
            
        # keep what you are doing
        register_handler_module(streamsyncuserapp)
```

This second part is given for information purposes only. It can take multiple forms.

```python
def init_handlers(handler_modules: Union[List[ModuleType], ModuleType]):
	app_process = get_app_process()
    if isinstance(handler_modules, ModuleType)
      handler_modules = [handler_modules]

    for module in handler_modules:
        _register_handler_module(app_process, module)	
    else:
        raise ValueError(f"Attempted to register a non-module object: {module}")

def _register_handler_module(app_process, module: ModuleType):
    """
    just a proposal, it can be anywhere (in class, inside AppProcess, ...)
    """
    pass

def _register_handler_function(app_process, func: Callable):
    pass
```

#### AppProcess as a kernel of streamsync application

**streamync loads and executes an application in AppProcess**. This component is responsible for loading the code of a streamsync application, processing messages received by the server, initiating and managing sessions.

``AppProcess.run`` corresponds to the main streamsync thread. Some examples of processing associated with AppProcess which clearly illustrates its level of responsibility

| functions                |                                                     |
| ------------------------ | --------------------------------------------------- |
| _execute_user_code       | load and execute the code of streamsync application |
| _get_user_functions      | liste the event handlers                            |
| _handle_component_update | update the code en mode édition                     |
| _handle_message          | process a message receive by streamsync server      |

image

*schema that illustrate the role of AppProcess regards to other components*

On the other hand, to work, a function like ``streamsync.init_handlers`` must be able to retrieve the context in which it executes, therefore the app process. **get_app_process is missing in the streamsync architecture**. 

A streamsync application exists because it is associated with an AppProcess. Today, it is an inaccessible aspect of the system and we work around it with the variables `initial_state`, `session_manager`, `base_cmc_tree`, ... These variables are indirect dependencies of the AppProcess (see my comment in the schema on SessionManager). This dependency is invisible but it exists thanks to the fork done in AppRunner._start_app_process. I spend several time to understand the magic why the session manager was not shared when mounting multi apps.

I would advocate introducing a **get_app_process function to allow registering event handlers** and removing some of the global variables `initial_state`, `session_manager`, `base_cmc_tree`, ... which are crutches.

**get_app_process is an "internal" function for streamsync maintainers and should be documented as such**. This function must be documented as part of the internal streamsync API and therefore excluded from the scope on which the team commits to maintaining backwards compatibility.

ramedina86:
Hey @FabienArcellier thanks for the insight, I fully agree with not relying on `streamsyncuserapp`.

I got used to `initial_state`, etc as global vars, and never saw them as something negative, but after reading your comment, it makes total sense. I wouldn't necessarily call it "magic" but it's a pattern that's far from elegant, so I fully support this change.

Please work with @mmikita95 to implement it.

src/streamsync/core.py
```diff
@@ -45,6 +48,20 @@
     StreamsyncFileItem,
 )
 
+if TYPE_CHECKING:
+    from streamsync.app_runner import AppProcess
+
+
+def get_app_process() -> 'AppProcess':
+    from streamsync.app_runner import AppProcess  # Needed during runtime
+    raw_process: BaseProcess = \
```
FabienArcellier, line 57: it's great to take that from the multiprocess directly I wasn't thinking of that.

src/streamsync/core.py
```diff
@@ -722,6 +739,93 @@ def call_frontend_function(self, module_key: str, function_name: str, args: List
...
 
+class EventHandlerRegistry:
+    """
+    Maps functions registered as event handlers from the user app's core
+    and external modules, providing an access mechanism to these maps.
+    """
+
+    HandlerArgsSequence = Sequence[str]
+    HandlerMetaEntry = Dict[str, Union[str, HandlerArgsSequence]]
+    HandlerEntry = Dict[str, Union[Callable, HandlerMetaEntry]]
```
FabienArcellier, line 750: Why are you permissive in the type declaration ? The format of the event looks highly specified. You should use TypedDict declaration

```python
{
    "callable": handler,
    "meta": {
        "name": access_name,
        "args": inspect.getfullargspec(handler).args
    }
}
```

src/streamsync/core.py
```diff
@@ -722,6 +739,93 @@ def call_frontend_function(self, module_key: str, function_name: str, args: List
...
+                fn_callable = getattr(module, fn_name)
+                if not fn_callable:
+                    continue
+                self.register_handler(fn_callable)
+        else:
+            raise ValueError(
+                f"Attempted to register a non-module object: {module}"
+                )
+
+    def find_handler(self, handler_name: str) -> Optional[Callable]:
```
FabienArcellier, line 800: I think this implementation is easier to understand

```python
def find_handler(self, handler_name: str) -> Optional[Callable]:
    if handler_name not in self.handler_map:
        return None
    
    return self.handler_map[handler_name]
```

src/streamsync/core.py
```diff
@@ -722,6 +739,93 @@ def call_frontend_function(self, module_key: str, function_name: str, args: List
...
+
+    def get_handler_meta(
+            self,
+            handler_name: str
+            ) -> "EventHandlerRegistry.HandlerMetaEntry":
+        if handler_name not in self.handler_map:
+            raise RuntimeError(f"Handler {handler_name} is not registered")
+        entry: EventHandlerRegistry.HandlerEntry = \
+            self.handler_map[handler_name]
+        if "meta" not in entry:
```
FabienArcellier, line 815: that should not happen ? it set inside `register_handler`.

mmikita95: I vaguely remembered adding this because mypy was upset 😄 but now, verifying it, seems that it is not the case. Good point, I'll fix this.

PR #374, opened 2024-04-03
feat: replace mode for UI manager

mmikita95:
Adds a `ui.refresh_with()` method that allows to clear existing children of a CMC and to attach new ones.


src/streamsync/core_ui.py
```diff
@@ -164,9 +164,13 @@ def clear_children(self, component_id: str) -> None:
             try:
                 self.delete_component(child.id)
             except UIError:
+                continue
```
FabienArcellier, line 167: We should do it before merging. I would propose something like that :

```python
logger = logging.getLogger("streamsync")
logger.warning("Failed to clear children of component with ID '{component_id}': {child.id} is a builder-managed component.")

File: src/streamsync/ui_manager.py [hunk 2]
```diff
@@ -69,6 +69,41 @@ def find(component_id: str) \
...
+    @staticmethod
+    def refresh_with(component_id: str):
+        """
+        Clears the existing children of a container component and sets it up to
+        accept new components. This method is designed to refresh the specified
+        container with new content specified in the subsequent block.
+
+        :param component_id: The unique identifier of the container component
+                             to be refreshed.
+        :type component_id: str
```
FabienArcellier, line 81: not necessary, the typing is already in the function signature

docs/docs/backend-driven-ui.md
```diff
@@ -62,6 +62,28 @@ with ui.find("column-container"):
...
+```
+
+This method also allows to clear children of a component:
+```python
+with ui.refresh_with("my-page"):
+    # Empties the page
+    pass
+```
+
+This method works only with CMCs exclusively and doesn't allow to modify BMCs, which will raise an `UIError`.
```
FabienArcellier, line 84: I think you remove the `UIError` parts. I would propose something like :+1: 

```
Only components created using backend are removed, builder component are preserved. You can't remove them programmatically. We recommand using the visibility flag to mask them.
```

docs/docs/backend-driven-ui.md
```diff
@@ -62,6 +62,28 @@ with ui.find("column-container"):
...
+
+This method also allows to clear children of a component:
+```python
+with ui.refresh_with("my-page"):
+    # Empties the page
+    pass
+```
+
+This method works only with CMCs exclusively and doesn't allow to modify BMCs, which will raise an `UIError`.  
+As well as with `find` method, it also raises a `RuntimeError` if it failes to find a referenced component.
```
FabienArcellier, line 85: I think it is `fails` instead of `failes`

FabienArcellier:
That looks ok for me

Peek 2024-04-08 08-05

FabienArcellier (approved):

src/streamsync/core_ui.py
FabienArcellier: currently I don't think we have. we would have to design one. I propose the pattern above because it may evolve easily and report logging error in a consistent way till we have better specification on logging functions / module.

mmikita95:
@ramedina86 I believe we're good to merge on this one. If you have any additional questions or concerns, I'll be happy to assist.

ramedina86:
CI failing, please fix.

mmikita95:
Good to go!

PR #386, opened 2024-04-15
fix: generation of documentation for components that fails on npm docs:build

FabienArcellier:
fix https://github.com/FabienArcellier/streamsync/issues/51


ramedina86:
Does this work on node.js?

FabienArcellier:
yes, I follow the same pattern than @raaymax

PR #395, opened 2024-04-22
fix: components are missing into documentation generation

FabienArcellier:


.github/workflows/ci-macos.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: macos-13
```
FabienArcellier, line 14: fail to start python 3.9 on macos 14 (M1)

The version '3.9' with architecture 'arm64' was not found for macOS 14.4.1.
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

FabienArcellier: other version are working fine : https://github.com/streamsync-cloud/streamsync/actions/runs/8795466464/job/24136566011

PR #402, opened 2024-04-25
docs: `init_handlers` usage

mmikita95:
Adding a section describing the usage of `init_handlers` method.


ramedina86:
Hey I don't think we've changed anything since then but can you confirm this is up to date?

PR #412, opened 2024-05-08
feat: cli deploy command

raaymax:


src/streamsync/command_line.py
```diff
@@ -31,11 +34,12 @@ def main():
...
-    _perform_checks(command, absolute_app_path, host, enable_remote_edit)
-    _route(command, absolute_app_path, port, host, enable_remote_edit)
+    _perform_checks(command, absolute_app_path, host, enable_remote_edit, api_key)
+    _route(command, absolute_app_path, port, host, enable_remote_edit, api_key)
 
-def _perform_checks(command: str, absolute_app_path: str, host: Optional[str], enable_remote_edit: Optional[bool]):
+def _perform_checks(command: str, absolute_app_path: str, host: Optional[str], enable_remote_edit: Optional[bool], api_key: Optional[str] = None):
     is_path_folder = absolute_app_path is not None and os.path.isdir(absolute_app_path)
 
     if command in ("run", "edit") and is_path_folder is False:
```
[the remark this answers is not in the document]
ramedina86: You'll have to modify the message too as in "is required to create or a deploy a Streamsync app"

src/streamsync/command_line.py
```diff
@@ -31,21 +34,29 @@ def main():
...
         logging.error("A path to a folder containing a Streamsync app is required. For example: streamsync edit my_app")
         sys.exit(1)
 
     if command in ("create") and absolute_app_path is None:
         logging.error("A target folder is required to create a Streamsync app. For example: streamsync create my_app")
         sys.exit(1)
 
+    if command in ("deploy") and api_key is None:
+        logging.info("An API key is required to deploy a Streamsync app.")
+        api_key = input("Enter your API key: ")
```
ramedina86, line 55: Nice, thanks for making that change. I think that'll display the token, right?

I found this which should work, can you test?

https://docs.python.org/3.9/library/getpass.html

ramedina86 (commented):
Looking good, just 1 comment

apps/default/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.10-bullseye
+RUN apt-get update -y && mkdir /app
+RUN apt-get install build-essential cmake python3-dev -y
+COPY . /app
+WORKDIR /app
+RUN pip3 install poetry
+RUN poetry config virtualenvs.create false
+RUN poetry install --only main
+ENTRYPOINT [ "streamsync", "run" ]
```
ramedina86, line 9: Can you change dockerfiles to be writer run, please?

src/writer/command_line.py
```diff
@@ -37,14 +41,32 @@ def main():
...
+    api_key = _get_api_key(command, api_key)
+    _route(command, absolute_app_path, port, host, enable_remote_edit, enable_server_setup_hook, template_name, api_key)
+
+def _get_api_key(command, api_key: Optional[str]) -> Optional[str]:
+    if command in ("deploy") and api_key is None:
+        env_key = os.getenv("WRITER_API_KEY", None)
+        if env_key is not None and env_key != "": 
+            return env_key
+        else:
+            logging.info("An API key is required to deploy a Streamsync app.")
```
ramedina86, line 56: Writer app, please can you do a search of `streamsync` to find any old references

apps/default/Dockerfile
raaymax: removed dockerfiles

ramedina86: Even better

PR #440, opened 2024-05-24
chore: verbose error for missing API key

mmikita95:
Added a descriptive error in case SDK client failed to initialize due to API key not being provided.

### DIFF
--- a/src/writer/ai.py
+++ b/src/writer/ai.py
@@ -3,9 +3,9 @@ from typing import Generator, Iterable, List, Literal, Optional, TypedDict, Unio
 
 from httpx import Timeout
 from writerai import Writer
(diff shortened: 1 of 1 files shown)


src/writer/ai.py
```diff
@@ -70,13 +71,20 @@ def __init__(self, token: Optional[str] = None):
...
-            # This is the default and can be omitted
-            api_key=token,
-        )
-
+        try:
+            self.client = Writer(
+                # This is the default and can be omitted
+                api_key=token,
+            )
+        except WriterError:
```
ramedina86, line 81: I don't see that we're actually checking the key, what happens if there's no internet connectivity, wouldn't this fail and tell you you don't have a key?

PR #446, opened 2024-05-27
feat: implement authentication workflow using basic auth

FabienArcellier:
For users who want to restrict access to their application for small application, I have implemented a simple basic auth restriction mechanism. 

*setup_server.py*
```python
auth = auth.BasicAuth(
    login=os.get('LOGIN'),
    password=os.get('PASSWORD'),
    delay_after_failure=1, # limit attempt when authentication fail (avoid brute force)
    block_user_after_failure=True # block the request for the delay
)
```

#### Brute force protection

A simple brute force protection is implemented by default. If a user fails to log in, the IP of this user is blocked.
Writer framework will ban the IP from either the `X-Forwarded-For` header or the `X-Real-IP` header or the client IP address.

When a user fails to log in, they wait 1 second before they can try again. This time can be modified by
modifying the value of `delay_after_failure`.

image


src/writer/auth.py
```diff
@@ -35,12 +39,106 @@ class Auth:
...
+
+        @asgi_app.middleware("http")
+        async def basicauth_middleware(request: Request, call_next):
+            import base64
+            client_ip = _client_ip(request)
+
+            try:
+                if client_ip in failed_attempts and time.time() - failed_attempts[client_ip] < self.delay_after_failure:
+                    remaining_time = int(self.delay_after_failure - (time.time() - failed_attempts[client_ip]))
+                    raise Unauthorized(status_code=429, message="Too Many Requests", more_info=f"You can try to log in every {self.delay_after_failure}s. Your next try is in {remaining_time}s.")
```
ramedina86, line 107: Hey, are we reusing `Unauthorized` for both? Shouldn't something like `ClientError` be more descriptive? I got confused with the 429 using the Unauthorized template

FabienArcellier: I decided to use the same exception because the authentication attempt is refused.

It could be an error from the client but also from the user. If he makes a mistake too many times, he reaches his maximum number of attempts in a given time. This can happen if the user persists so I prefer that they receive an explicit error message. A bit like when the user locks their phone because they entered the wrong PIN too many times.

I can create a ClientError exception which Unauthorized will inherit. Would that be more relevant?

ramedina86: Makes sense, let's leave it as is.

PR #462, opened 2024-06-13
Annotated Text component

polymorpheuz:


src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+}
+
+const defaultShortNameDictionary = {
+	verb: "Verb",
+	adjective: "Adj",
+	noun: "Noun",
+	pronoun: "Pronoun",
+};
+
+function getShortName(name: string) {
```
polymorpheuz, line 268: Not sure if we even need it. In our reference package they do things like that.

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+		source.clipboardData?.setData("text/plain", text);
+		source.clipboardData?.setData("Text", text); // IE mimetype
+	}
+
+	if (html) {
+		source.clipboardData?.setData("text/html", html);
+	}
+}
+
+function copyToClipboard({
```
polymorpheuz, line 216: I believe these two copy functions should travel someplace else but I have no understanding where to.

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+	{ h: -12, s: 0, l: 29 },
+	{ h: 28, s: -20, l: 24 },
+	{ h: -61, s: -95, l: 25 },
+	{ h: -173, s: 0, l: 16 },
+	{ h: -228, s: 0, l: 22 },
+	{ h: 69, s: 0, l: 25 },
+	{ h: 70, s: 0, l: 29 },
+];
+
+let currentSteps = [...COLOR_STEPS];
```
polymorpheuz, line 116: You might wonder why am I doing that. For the tags it's perfectly fine to have the same color for for different tags. Here it's critical to not repeat the colors. More below.

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+
+	// If we run out of colors, reset the list
+	if (currentSteps.length === 0) {
+		currentSteps = [...COLOR_STEPS];
+	}
+
+	const colorStep = calculateColorStep(s, currentSteps.length);
+	const colorData = currentSteps[colorStep];
+
+	subjectColorCache[s] = colorData;
```
polymorpheuz, line 176: So we make sure to cache the initial allocation here

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+	currentSteps.splice(colorStep, 1);
```
polymorpheuz, line 177: ..and we get rid of that color so the random pick won't set us up to have the same color for another subject

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+		subjectColorCache = {};
+		lastSeed = fields.seed.value;
+	}
+
+	if (subjectColorCache[s]) {
+		return generateColorCss(baseColor, subjectColorCache[s]);
+	}
+
+	// If we run out of colors, reset the list
+	if (currentSteps.length === 0) {
```
polymorpheuz, line 169: This is a safeguard in order to keep it operational even if we ran out of colors. We shouldn't allow that though and it's better to add like +10-20 color to what we have now.

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+		} finally {
+			document.body.removeChild(textarea);
+			document.removeEventListener("copy", copyListener, false);
+		}
+	}
+
+	return false;
+}
+
+const defaultShortNameDictionary = {
```
ramedina86, line 261: Let's remove this dict, people can enter the whole thing

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+	border-radius: 0.5rem;
+	padding: 0.08rem 0.5rem;
+	overflow: hidden;
+	line-height: 1;
+	vertical-align: middle;
+}
+
+.annotation-subject {
+	display: inline-flex;
+	font-size: 0.6rem;
```
ramedina86, line 301: Let's use px instead of rem

I use rem for font-sizes, but looking to change to px too

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+	const colorStep = calculateColorStep(s, currentSteps.length);
+	const colorData = currentSteps[colorStep];
+
+	subjectColorCache[s] = colorData;
+	currentSteps.splice(colorStep, 1);
+
+	return generateColorCss(baseColor, colorData);
+}
+
+function copyText(arr: string[]) {
```
ramedina86, line 182: Let's move copy functionality and the bar to `base/` in a component called `BaseControlBar.vue` (or similar).

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+	{ h: -228, s: 0, l: 22 },
+	{ h: 69, s: 0, l: 25 },
+	{ h: 70, s: 0, l: 29 },
+];
+
+let currentSteps = [...COLOR_STEPS];
+let subjectColorCache = {};
+let lastSeed = fields.seed.value;
+
+function generateColorCss(
```
ramedina86, line 120: Feel free about moving to `colorTransformers` or similar (if you want).

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+	{ h: -61, s: -37, l: 16 },
+	{ h: -2, s: 0, l: 24 },
+	{ h: -12, s: 0, l: 29 },
+	{ h: 28, s: -20, l: 24 },
+	{ h: -61, s: -95, l: 25 },
+	{ h: -173, s: 0, l: 16 },
+	{ h: -228, s: 0, l: 22 },
+	{ h: 69, s: 0, l: 25 },
+	{ h: 70, s: 0, l: 29 },
+];
```
ramedina86, line 114: Think about how to extend to support more colors, maybe consider discussing this with Miffy. Right now, I'm using the Tag colors as specified in the design system.

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,339 @@
...
+		name: "Annotated text",
+		description: "Shows text with annotations",
+		category: "Content",
+
+		fields: {
+			text: {
+				name: "KeyValue",
+				type: FieldType.Object,
+				desc: "Value array with text/annotations. Must be a JSON string or a state reference to an array.",
+				default: `["This ",["is", "Verb"]," some ",["annotated", "Adjective"],    ["text", "Noun"]," for those of ",["you", "Pronoun"]," who ",["like", "Verb"]," this sort of ",["thing", "Noun"],". ","And here's a ",["word", "", "#faf"]," with a fancy background but no label."]`,
```
ramedina86, line 50: Let's go for `init` rather than `default`

ramedina86:
Forgot to mention, we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

src/ui/src/core_components/content/CoreAnnotatedText.vue
ramedina86: @polymorpheuz We may choose to have a file for color transformations, since they're similar across tags and annotated text

ramedina86:
Hey @polymorpheuz can you add the component to `templateMap.ts`? So that it's linked (unsure how you're developing right now)

@FabienArcellier this is a very interesting component if you want to have a look

src/ui/src/core_components/base/BaseControlBar.vue
```diff
@@ -0,0 +1,110 @@
...
+		textarea.select();
+
+		try {
+			return document.execCommand("copy"); // Security exception may be thrown by some browsers.
+		} catch (ex) {
+			return false;
+		} finally {
+			document.body.removeChild(textarea);
+			document.removeEventListener("copy", copyListener, false);
+		}
```
madeindjs, line 79: `document.execCommand` is deprecated API ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)). You should use [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) like this

```ts
await navigator.clipboard.writeText(text);
```

You can even use this solution [with this polyfill](https://github.com/lgarron/clipboard-polyfill) to support old browsers

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,248 @@
...
+	writer: {
+		name: "Annotated text",
+		description: "Shows text with annotations",
+		category: "Content",
+
+		fields: {
+			text: {
+				name: "KeyValue",
+				type: FieldType.Object,
+				desc: "Value array with text/annotations. Must be a JSON string or a state reference to an array.",
```
madeindjs, line 39: It needs to be a an array, no ?
```suggestion
				desc: "Value array with text/annotations. Must be a JSON Array as string or a state reference to an array.",
```

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,248 @@
+<template>
+	<div class="AnnotatedText">
+		<span v-for="(content, i) in fields.text.value" :key="content + i">
```
madeindjs, line 3: `content` can be a number and then you can have duplicated key. I suggest to cast it to be safe
```suggestion
		<span v-for="(content, i) in fields.text.value" :key="String(content) + i">
```

src/ui/src/core_components/content/CoreAnnotatedText.vue
```diff
@@ -0,0 +1,248 @@
...
+	{ h: -61, s: -95, l: 25 },
+	{ h: -173, s: 0, l: 16 },
+	{ h: -228, s: 0, l: 22 },
+	{ h: 69, s: 0, l: 25 },
+	{ h: 70, s: 0, l: 29 },
+];
+
+let currentSteps = [...COLOR_STEPS];
+let subjectColorCache = {};
+let lastSeed = fields.seed.value;
```
madeindjs, line 108: Are you sure you don't need to make them reactive ? 
```suggestion
let currentSteps = ref([...COLOR_STEPS]);
let subjectColorCache = ref({});
let lastSeed = ref(fields.seed.value);
```

If its not refs, Vue.js can't track them and might not re-render on update ([doc](https://vuejs.org/guide/essentials/reactivity-fundamentals.html#why-refs))

polymorpheuz: I don't think that we need reactivity in this case. I am following `CoreTags` that seem to work in this case just as fine.

src/ui/src/core_components/base/BaseControlBar.vue
```diff
@@ -0,0 +1,110 @@
...
+			Copy
+		</button>
+	</div>
+</template>
+
+<script setup lang="ts">
+const props = defineProps<{
+	copyRawContent?: string;
+	copyStructuredContent?: string;
+}>();
```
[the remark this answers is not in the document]
polymorpheuz: Maybe it's a broader question on how we actually define props in the project. In this case, I was following how the other components defining props. My take is to stick to one approach until we decide to go with another.

Also, I'm not sure if there's practical use of runtime checks in this case. The bar is used by us as a building block, so it's not a blackbox for a general user to leave hints to.

@ramedina86 @FabienArcellier What do you think?

src/ui/src/core_components/base/BaseControlBar.vue
polymorpheuz: Thanks for pointing this out!

It's [not deprecated](https://stackoverflow.com/a/78427138) in a traditional sense. The drama around it is long and could take 5 at least more years. I went with it since I don't see it being deprecated anytime soon and it just works for any browser. It just works for any content format, so the component is easily extensible.

However, it's a nice discussion point. I couldn't find a doc on what browsers do we actually support. If we are not hitting limitations with these [constraints](https://github.com/lgarron/clipboard-polyfill?tab=readme-ov-file#compatibility), I have no objections on using `navigator.clipboard`.

cc @ramedina86 @FabienArcellier

src/ui/src/core_components/content/CoreAnnotatedText.vue
polymorpheuz: @ramedina86 I've asked Miffy a while ago on the additional colors and she pointed to the other color groups. The problem is that they don't match 😅 I was waiting for Pete being in the office to ambush him but he keeps avoiding me so far.

It seems like it's either we're going with 10 colors or an LLM is out best bet 🤣

ramedina86:
@polymorpheuz Have you looked into why CI isn't passing?

src/ui/src/core_components/base/BaseControlBar.vue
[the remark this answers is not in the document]
ramedina86: I think this is a good suggestion but in line with @polymorpheuz I'd like to decide on the standard and then @madeindjs can migrate all in one block :)

src/ui/src/core_components/content/CoreAnnotatedText.vue
ramedina86: Yes, let's not use reactivity that we don't need.

PR #467, opened 2024-06-17
docs: page for docker

raaymax:


docs/docs/deploy-with-docker.md
```diff
@@ -17,72 +74,105 @@ A Dockerfile is a file with instructions that tell Docker how to build your imag
...
 
-```sh
-docker build . -t my_streamsync_app
-```
+1. **Navigate to your project directory**:
+   ```sh
+   cd path/to/your/app
+   ```
 
-::: warning Platform considerations
```
ramedina86, line 46: Hey why was this removed, has it stopped being a problem?

docs/docs/deploy-with-docker.md
```diff
@@ -1,14 +1,71 @@
...
+   ```
+   This command will update your `pyproject.toml` file with `writer` as a dependency and install it.
+
+4. **Install Dependencies**: Ensure all dependencies are installed in your virtual environment.
+   ```sh
+   poetry install
+   ```
+   This will create a virtual environment (if it doesn't already exist) and install all dependencies listed in `pyproject.toml`.
+
+### What is Docker and Why Use It?
```
ramedina86, line 58: @raaymax  @anant-writer I don't think we need this section. If we're targeting people who have no idea what Docker is, we'll need to be much more explicit about everything. We're transitioning from some high-level features to "Creating a Dockerfile" without discussing what an image is, what a Dockerfile is.

ramedina86: Also, this whole thing is inside a section called "Creating a Docker image", we really need to rethink the subtitle if we're going so general.

docs/docs/deploy-with-docker.md
```diff
@@ -1,14 +1,71 @@
...
 ## Creating a Docker image
 
 ### Setting up
 
 - Make sure you have Docker installed.
 - Open a terminal and navigate to your app's folder.
-- Create a `pyproject.toml` using `poetry init` and install `streamsync` using `poetry add streamsync`
+- Create a `pyproject.toml` using `poetry init` and install `writer` using `poetry add writer`
+
+### What is Poetry and Its Purpose?
```
ramedina86, line 13: @raaymax @anant-writer I'd phrase more like "Poetry is [...], and Writer Framework uses it to [...]." It needs to be clear that it isn't an option.

docs/docs/deploy-with-docker.md
```diff
@@ -1,14 +1,71 @@
...
+You can use Docker to deploy Writer anywhere. If you're an experienced Docker user, you may want to go straight to the provided Dockerfile.
 
 ## Creating a Docker image
 
 ### Setting up
 
 - Make sure you have Docker installed.
 - Open a terminal and navigate to your app's folder.
-- Create a `pyproject.toml` using `poetry init` and install `streamsync` using `poetry add streamsync`
+- Create a `pyproject.toml` using `poetry init` and install `writer` using `poetry add writer`
```
ramedina86, line 11: `pyproject.toml` is included in the created app. Shouldn't we use that?

ramedina86 (changes requested):
@anant-writer @raaymax Let's rethink this

PR #469, opened 2024-06-19
feat: Added redeploy confirmation if different app was already deployed using same api key

raaymax:


FabienArcellier:
In this scenario, wouldn't it be simpler to use the absolute path of the folder that contains the application as a reference?

The only benefit of using app id is to support directory rename. If you go further in this direction, I would recommand to add it on the app provisionning or in the writer framework code, not in the deployment code.

alternate propsition that does not change the ui.json :

* have a file `deployments.json`

```json
{
  "<token>": "<absolute path of project>",
  "<token2>": "<absolute path of project 2>"
}
```

* raise a warning when a token is already used for a path that is different from the previous one

ramedina86:
Good to close this?

PR #471, opened 2024-06-19
use middleware to intercept event handlers (global event, database management, ...)

FabienArcellier:
This PR implements the concept of middleware. Middleware is a function that intercepts all events. You can use it to implement session management for a database, code an extension to streamsync which globally modifies the operation of certain events, ...

Peek 2024-06-19 23-20

```python
# MIDDLEWARES
@wf.middleware()
def count_events(state):
    state['counter_event'] += 1
    yield

@wf.middleware()
def count_specific_events(state, context):
    event_type = context.get('event')
    if event_type == 'wf-click':
        state['counter_event_click'] += 1

    if event_type == 'wf-option-change':
        state['counter_event_select'] += 1

    yield
```

### Usecase

**middleware facilitates cross-functional operations such as database transaction management**. Opening a session at the start of a query is a pattern that improves the use of a database. The session links the transaction to the request. If processing fails, the transaction is rolled back. If it succeeds, the transaction is successful.

```python
import writer as wf

engine = sa.create_engine("postgresql://admin:admin@localhost/Adventureworks", echo=True)
Session = sessionmaker(engine)

@wf.middleware
def manage_sqlalchemy_session():
    with Session() as session:
        session.begin()
        try:
            yield
            session.commit()
        except:
            session.rollback()
            raise
```

**middleware facilitates cross-functional operations such as exception tracking**.

```python
import writer as wf

@wf.middleware
def manage_sentry_capture(state, payload):
    with sentry_sdk.configure_scope() as scope:
        scope.set_tag("site", state['site'])
        try:
            yield
        except Exception as exception:
            sentry_sdk.capture(exception)
            raise
        finally:
            scope.clear()
```


src/writer/__init__.py
```diff
@@ -128,3 +129,34 @@ def init_handlers(handler_modules: Union[List[ModuleType], ModuleType]):
...
+    >>> import writer as wf
+    >>>
+    >>> @wf.middleware()
+    >>> def my_middleware(state, payload, session):
+    >>>     state['processing'] += 1
+    >>>     yield
+    >>>     state['processing'] -= 1
+    """
+    def inner(func):
+        # enregistre la fonction en tant que middlewares
```
ramedina86, line 157: English please 🏴󠁧󠁢󠁥󠁮󠁧󠁿 (or remove)

src/writer/tests.py
```diff
@@ -0,0 +1,21 @@
+def invoke_event_handler(event_handler, **kwargs):
+    """
+    Invokes an event handler in the same way that streamsync would.
```
ramedina86, line 3: Please change streamsync to Writer Framework

FabienArcellier: this code was not necessary. I removed it

src/writer/core.py
```diff
@@ -737,6 +743,36 @@ def call_frontend_function(self, module_key: str, function_name: str, args: List
             "args": args
         })
 
+class MiddlewareExecutor():
```
FabienArcellier, line 746: write a description about the role of this class.

```python
    """
    A MiddlewareExecutor executes middleware in a controlled context. It allows writer framework
    to manage different implementations of middleware.

    Case 1 : A middleware is a generator, then run before and after code

    >>> @wf.middleware()
    >>> def my_middleware():
    >>>     print("before event handler")
    >>>     yield()
    >>>     print("after event handler")

    Case 2 : A middleware is just a function, then run the function before

    >>> @wf.middleware()
    >>> def my_middleware():
    >>>     print("before event handler")
    """
```

src/writer/core.py
```diff
@@ -1331,51 +1357,43 @@ def _call_handler_callable(self, event_type, target_component, instance_path, pa
...
-            if is_async_handler:
-                result, captured_stdout = self._async_handler_executor(callable_handler, arg_values)
-            else:
-                result, captured_stdout = self._sync_handler_executor(callable_handler, arg_values)
+        captured_stdout = None
+        with core_ui.use_component_tree(self.session.session_component_tree), \
+            contextlib.redirect_stdout(io.StringIO()) as f:
+            middlewares_executors = current_app_process.middleware_registry.executors(writer_args)
+
+            handle_with_middlewares_executor(middlewares_executors, callable_handler, writer_args)
```
FabienArcellier, line 1387: the result of event handler was not set in result.

```python
middlewares_executors = current_app_process.middleware_registry.executors()
result = handle_with_middlewares_executor(middlewares_executors, callable_handler, writer_args)
```

src/writer/app_runner.py
```diff
@@ -255,6 +271,13 @@ def _handle_message(self, session_id: str, request: AppProcessServerRequest) ->
                 payload=self._handle_state_enquiry(session)
             )
 
+        if type == "stateContent":
+            return AppProcessServerResponse(
+                status="ok",
+                status_message=None,
+                payload=self._handle_state_full(session)
```
FabienArcellier, line 278: rename into `_handle_state_content`

PR #474, opened 2024-06-25
docs: improve documentation to use writer instead of streamsync

### LINKED ISSUE #473: Problem with StreamSyncState
This is the problem when using writer.StreamSyncState:  
  
class AppSchema(wf.StreamsyncState):
AttributeError: module 'writer' has no attribute 'StreamsyncState'

FabienArcellier:
There are still dependencies on streamsync in the documentation. I fixed the one I found.

fix #473

### DIFF
--- a/docs/components/component_page.mdx.tpl
+++ b/docs/components/component_page.mdx.tpl
@@ -71,4 +71,4 @@ A function, in this example `handle_event`, should be implemented in your code t
 
 ## Reference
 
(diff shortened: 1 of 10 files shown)


ramedina86:
Hey does this still need to be merged? I thought Anant took care of the string replacements?

FabienArcellier:
I think you can. It looks like some parts still need to be updated.

image

PR #480, opened 2024-06-27
fix: enable the server setup hook by default in a custom server

FabienArcellier:


ramedina86:
Does this exclude `writer edit`? I think it should be deactivated for `writer edit` but activated by default for run even with custom server.

FabienArcellier:
> Does this exclude `writer edit`? I think it should be deactivated for `writer edit` but activated by default for run even with custom server.

This excludes the writer edit command. This is handled in calling code. 

On the other hand, if you mount the application in edition in a multi app application, the edition will have authentication activated. You will have to deactivate it explicitly.

PR #481, opened 2024-06-28
fix: `BaseSelect` handle `number` value. WF-28

madeindjs:
A quick fix to handle number values in `BaseSelect` (used in `MultiSelect` & `SelectInput`).

[Screencast from 2024-06-28 09-27-30.webm](https://github.com/writer/writer-framework/assets/11815139/69ac333a-f248-44ad-9ff0-2257c5b2f291)

I also did a small refactoring on props definition, because using object is more robust (it does a runtime check)

### DIFF
--- a/src/ui/src/core_components/base/BaseSelect.vue
+++ b/src/ui/src/core_components/base/BaseSelect.vue
@@ -93,21 +93,21 @@
 </template>
 
 <script setup lang="ts">
-import { computed, Ref, toRefs } from "vue";
-import { nextTick } from "vue";
-import { ref } from "vue";
-import { watch } from "vue";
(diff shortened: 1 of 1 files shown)


FabienArcellier:
I would advocate to open a issue to document slightly the bug. You can directly reproduce the video.

src/ui/src/core_components/base/BaseSelect.vue
```diff
@@ -93,21 +93,21 @@
...
 
-const props = defineProps<{
-	baseId: string;
-	activeValue: any;
-	options: Record<string, string>;
-	maximumCount: number;
-	mode: "single" | "multiple";
-	placeholder?: string;
-}>();
+const props = defineProps({
```
FabienArcellier, line 100: It may be something we have to generalize on next PR that change a component.

PR #482, opened 2024-06-29
fix: Hide `CoreText` in preview if empty value. WF-20

madeindjs:
Simply hide the `CoreText` using `v-if` in preview mode if the value is `emptyText`. We still display the component in edit mode.

[Screencast from 2024-06-29 14-37-01.webm](https://github.com/writer/writer-framework/assets/11815139/12b438b5-dc39-452c-885e-e996b9e8ddfb)

### DIFF
--- a/src/ui/src/core_components/base/BaseEmptiness.vue
+++ b/src/ui/src/core_components/base/BaseEmptiness.vue
@@ -1,5 +1,5 @@
 <template>
-	<div v-if="displayEmpty" class="BaseEmptiness">
+	<div class="BaseEmptiness">
 		<div class="content">
 			<div class="title">
 				<h3>Empty {{ definition.name }}</h3>
@@ -9,7 +9,6 @@
 			</div>
 		</div>
 	</div>
-	<slot v-else-if="displayContent" />
 </template>
 
 <script setup lang="ts">
@@ -19,7 +18,6 @@ import injectionKeys from "../../injectionKeys";
 const props = defineProps({
 	componentId: { type: String, required: true },
 	message: { type: String, required: false, default: undefined },
-	isEmpty: { type: Boolean },
 });
 
 const wf = inject(injectionKeys.core);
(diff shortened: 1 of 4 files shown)


FabienArcellier:
We discuss about generalizing the way we manage this sort of state. We will try to use EmptinessLabel to encapsulate the behavior.

* make a list of components to change

madeindjs:
Hello @FabienArcellier & @ramedina86 , 

I'm trying to implement a general behavior to handle empty state for CoreText & CoreMessage . As @FabienArcellier  pointed, we might want to have a general way to do this.

I succeded to implement a `EmptinessState` component who does the job (https://github.com/writer/writer-framework/pull/482/commits/5ef475be9fd4be7a72d79fc6e063005a22ac0606) but it takes some space in preview mode (as we still render an empty comp).

[Screencast from 2024-07-16 00-24-36.webm](https://github.com/user-attachments/assets/09f875b2-0570-41da-a429-6530438628df)

I saw empty state of sidebar / Header / etc.. use a different logic. It uses `ChildlessPlaceholder` through ComponentProxy.

https://github.com/writer/writer-framework/blob/df6fa8d260ef8fd8cc8a50173b36e6943950ae43/ui/src/renderer/ComponentProxy.vue#L37-L42

So I wonder if we want the same mechanism or if it's a bit overkill for the need and the `v-if` is enough (https://github.com/writer/writer-framework/pull/482/commits/ba05e5045367a18b69ba6b10f82df38438dd6af3)

WDTY?

FabienArcellier:
> Hello @FabienArcellier & @ramedina86 ,
> 
> I'm trying to implement a general behavior to handle empty state for CoreText & CoreMessage . As @FabienArcellier pointed, we might want to have a general way to do this.
> 
> I succeded to implement a `EmptinessState` component who does the job ([5ef475b](https://github.com/writer/writer-framework/commit/5ef475be9fd4be7a72d79fc6e063005a22ac0606)) but it takes some space in preview mode (as we still render an empty comp).
> 
>  Screencast.from.2024-07-16.00-24-36.webm 
> I saw empty state of sidebar / Header / etc.. use a different logic. It uses `ChildlessPlaceholder` through ComponentProxy.
> 
> https://github.com/writer/writer-framework/blob/df6fa8d260ef8fd8cc8a50173b36e6943950ae43/ui/src/renderer/ComponentProxy.vue#L37-L42
> 
> So I wonder if we want the same mechanism or if it's a bit overkill for the need and the `v-if` is enough ([ba05e50](https://github.com/writer/writer-framework/commit/ba05e5045367a18b69ba6b10f82df38438dd6af3))
> 
> WDTY?

For the `empty state of sidebar / Header / etc`, I would stick to the current logic. The v-if looks fine.

ramedina86:
@madeindjs Those are two different things

Childless placeholder: Automatically applied to any component that can contain children but doesn't

The other one (what you've called `EmptinessState`) should be used for when rendering the component would yield undesirable results due to lack of configuration/content.

PR #484, opened 2024-07-01
fix: handle template in `BuilderFieldsKeyValue`. WF-31

madeindjs:
The issue was `BuilderFieldsKeyValue` didn't interpret the template syntax and try to `JSON.parse` it when navigating between "Static List" and "JSON".

I fixed it relying on `useEvaluator().getEvaluatedFields` which take care of getting the current data and parsing template. It simplify a lot the logic of the component!

It just required that I get and parse `instancePath` in the parent.

Now everything works as expected, if user modify a template value in "Static list" mode, the template is overwiten by the plain JSON value. We still handle the "undo/redo" as expected.

[Screencast from 2024-07-01 23-45-17.webm](https://github.com/writer/writer-framework/assets/11815139/4d831d32-1c40-44f3-82b6-b34c06ed14d8)

### DIFF
--- a/src/ui/src/builder/BuilderFieldsKeyValue.vue
+++ b/src/ui/src/builder/BuilderFieldsKeyValue.vue
@@ -75,29 +75,33 @@
 
 <script setup lang="ts">
 import {
+	PropType,
+	Ref,
+	computed,
+	inject,
 	nextTick,
 	onMounted,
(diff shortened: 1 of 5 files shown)


ramedina86:
Beautiful work! Love when things become simpler

PR #486, opened 2024-07-05
feat(ui): introduce `CoreColorInput`. WF-9

madeindjs:
Create a new component `CoreColorInput` using a sub-component `BaseInputColor`.

The component accept a list of predefined colors using `datalist` (see [doc](https://developer.chrome.com/blog/datalist-for-range-color-inputs-offer-some-default-choices/#inputtypecolor))

[Screencast from 2024-07-09 22-10-29.webm](https://github.com/writer/writer-framework/assets/11815139/0af2c953-85fe-4fb4-a4e6-b9af5da80dc2)


FabienArcellier:
The implementation look promising but I fails to reproduce the [video on review environment](https://streamsync-dev-pr486.osc-fr1.scalingo.io/default/).

FabienArcellier:
* set a field `color list` (json array ?)
* separate the label for the width of color picker
* use different component between builder and component
* for later, set existing color other than default one at the end, sort by occurence
* for later, for history parsing, use field type color instead of color matching

madeindjs:
@FabienArcellier & @ramedina86 , I implemented the changes. Here the new screenshot

[Screencast from 2024-07-09 22-10-29.webm](https://github.com/writer/writer-framework/assets/11815139/0af2c953-85fe-4fb4-a4e6-b9af5da80dc2)

I have a small doubt regarding predifined colors because:

1. we use `FieldType.Object` since we don't have array field, so we have [to check that the JSON is an array](https://github.com/writer/writer-framework/blob/bb8dc80ab1d552cfc3ff57957da27d27fcef2be3/src/ui/src/core_components/input/CoreColorInput.vue#L82-L84) and it's not convenient for the user to edit it
2. The browser accept only a list of hexadecimal colors (see [doc](https://developer.chrome.com/blog/datalist-for-range-color-inputs-offer-some-default-choices/))
3. It's just a suggestion palette, the user can use the color picker to select whatever he wants

Let me know what you think about it

src/ui/src/core_components/input/CoreColorInput.vue
```diff
@@ -0,0 +1,100 @@
...
+	category: "Input",
+	fields: {
+		label: {
+			name: "Label",
+			init: "Input Label",
+			type: FieldType.Text,
+		},
+		colorList: {
+			name: "Color List",
+			desc: "List of predefined colors",
```
FabienArcellier, line 45: `list of predefined colors in hexadecimal format`

FabienArcellier: what is the benefits of the event `wf-change-finish` ? I feel like the `wf-change` event covers all cases.

image

FabienArcellier:
Thanks It seems good. An improvement for later would be to add a `BuilderFieldsValueList` component which takes inspiration from `BuilderFieldsKeyValue` to manipulate a proposal list and provide preformatted Json.

src/ui/src/core_components/input/CoreColorInput.vue
madeindjs: @FabienArcellier I had the same question when seing this event on `CoreTextInput` ([here](https://github.com/writer/writer-framework/blob/6f7a03c8dce216adfe27bb0809e68045e0ed2f4b/src/ui/src/core_components/input/CoreTextInput.vue#L74-L82)). The difference is the HTML event tracked

https://github.com/writer/writer-framework/blob/bb8dc80ab1d552cfc3ff57957da27d27fcef2be3/src/ui/src/core_components/base/BaseInputColor.vue#L8-L9

- `@input` / `wf-change` is triggered when the value change in real time (like user type a letter or slide the color picker)
- `@change` / `wf-change-finish` is trigger when user finish to write and click outside (lose focus)

Thus it's better to have both because there are some situations where you want to execute the action once the user finished his changes

src/ui/src/core/index.ts
```diff
@@ -594,6 +594,7 @@ export function generateCore() {
 		getSessionTimestamp,
 		getUserState,
 		isChildOf,
+		components: computed(() => Object.values(components.value)),
```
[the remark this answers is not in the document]
ramedina86: You can use `getComponents` without a parent for that though.

PR #487, opened 2024-07-06
chore(ui): use local fonts instead of CDN. WF-27

madeindjs:
Use [Fontsource](https://fontsource.org/) with NPM instead of relying on https://fonts.google.com

Try to spot the difference (I don't see any :monocle_face: )

| before | CDN |
|--|--|
| Screenshot 2024-07-06 at 22-36-55 Hello Writer Framework Builder | Screenshot 2024-07-06 at 22-36-36 Hello Writer Framework Builder |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2587,6 +2587,16 @@
 			"dev": true,
 			"license": "MIT"
 		},
+		"node_modules/@fontsource-variable/material-symbols-outlined": {
+			"version": "5.0.34",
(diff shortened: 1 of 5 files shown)


ramedina86:
Hey, looks good, I didn't know about Fontsource. Neat to be able to manage via npm. 

Can we look into only using local if CDN fails?

https://stackoverflow.com/questions/7383163/how-to-fallback-to-local-stylesheet-not-script-if-cdn-fails

FabienArcellier:
I love it. Everything is served through the local server. The cache is used properly on the browser, so I would advocate to keep local diffusion instead of using cdn.

(left: dev version, right: version in the PR)
image

(cache usage, the file is only downloaded once)
image

PR #490, opened 2024-07-15
feat: SDK Knowledge Graphs / Files support

mmikita95:
- `SDKWrapper` class for dynamic property access of SDK objects that are expected to be used directly in Framework code.
- `Graph` class to wrap `writerai.types.Graph` with methods for property retrieval, updates, and file operations.
- `File` class to wrap `writerai.types.File`.
- Helper functions for graph and file operations (create, retrieve, list, delete, upload).


ramedina86:
@mmikita95 can you pin the version of the SDK to `0.5.0` in the toml please?

PR #494, opened 2024-07-18
feat(ui): improve `CoreFileInput`

madeindjs:
The main issue of `CoreFileInput` was that, when it re-render, the files selected label is cleared and there is no way to set it again.

To prevent that, the easiest way is to hide the `<input type="file" />` and display instead a custom button and a label.

This unlock a cool possibility to create the link of the file and let the user download the file.

[Screencast from 2024-07-18 23-00-18.webm](https://github.com/user-attachments/assets/3c054794-60e6-4999-82f7-0eb947212397)


ramedina86:
I like the logic of this one but I think the button needs to be changed.

Can we try using a plain`WdsButton`? By plain I mean `separatorColor` as border, then transparent or `containerBackgroundColor` for the background, `primaryTextColor` for the text.

From the design system:
 
Screenshot 2024-07-30 at 13 28 33

madeindjs:
@ramedina86 , do you think we can try to reuse the [`WdsButton`](https://github.com/madeindjs/writer-framework/blob/1148fc4a7eba17bb41154a7fc5ec48904cc76d2e/src/ui/src/wds/WdsButton.vue) and pass CSS var to customize it ?

It seems a bit hard for now though, the border and background can't be separated for instance

https://github.com/writer/writer-framework/blob/37b25b1bafcddbbe0909ca9482620273ad1a9fb4/src/ui/src/wds/WdsButton.vue#L15-L16

ramedina86:
@madeindjs please implement a `variant` prop, we'll need it anyway. If you check WDS we have several variants, so it makes sense to make it user-configurable for other buttons.

madeindjs:
@ramedina86 , It's done

ramedina86:
Nice, looks really good!

PR #496, opened 2024-07-24
feat: trigger a calculated property on mutation

FabienArcellier:
implement https://github.com/FabienArcellier/writer-framework/issues/37

Peek 2024-08-01 08-09

### Simple subscription

```python
def _increment_counter(state):
    state['my_counter'] += 1

_state = wf.init_state({"a": 1, "my_counter": 0})
_state.subscribe_mutation('a', _increment_counter)

_state['a'] = 2 # trigger _increment_counter mutation
```

### Subscription on state with strong type

```python
class MyState(wf.WriterState):
  counter: int
  total: int

def cumulative_sum(state: MyState):
  state.total += state.counter

initial_state = ss.init_state({
    "counter": 0,
    "total": 0
}, schema=MyState)

initial_state.subscribe_mutation('counter', cumulative_sum)

initial_state['counter'] = 2 # trigger the function cumulative_sum
```

### Calculated property

```python
class MyState(wf.WriterState):
    counter: int

    @wf.property('counter')
    def counter_str(self) -> str:
        return str(self.counter)

state = wf.init_state({'counter': 0}, MyState)

state.counter = 2 # trigger mutation for +counter_str 
```

### Subscription with an event handler

```python
def _increment_counter(state: RootState, context: dict, payload: dict, session: dict, ui: wf.WriterUIManager):
	if context['event'] == 'mutation' and context['mutation'] == 'a':
		if payload['previous_value'] < payload['new_value']:
			state['my_counter'] += 1

state = wf.init_state({
	"a": 1, 
	"my_counter": 0,
})

state.subscribe_mutation('a', _increment_counter)
```

### In preparation of dot support expression

```python
def cumulative_sum(state):
    state['total'] += state['a.b']

initial_state = wf.init_state({
    "a.b": 0,
    "total": 0
})

initial_state.subscribe_mutation('a\.b', cumulative_sum)
```

### DIFF
--- a/src/writer/__init__.py
+++ b/src/writer/__init__.py
@@ -17,6 +17,9 @@ from writer.core import (
     session_manager,
     session_verifier,
 )
+from writer.core import (
+    writerproperty as property,
+)
 from writer.ui import WriterUIManager
 
 VERSION = importlib.metadata.version("writer")
@@ -94,6 +97,7 @@ def init_state(raw_state: Dict[str, Any], schema: Optional[Type[S]] = None) -> U
         raise ValueError("Root schema must inherit from WriterState")
 
     _initial_state: S = new_initial_state(concrete_schema, raw_state)
+
     return _initial_state
 
 
--- a/src/writer/core.py
+++ b/src/writer/core.py
@@ -18,7 +18,6 @@ import traceback
 import types
 import urllib.request
 from abc import ABCMeta
-from functools import partial, wraps
 from multiprocessing.process import BaseProcess
 from types import ModuleType
 from typing import (
@@ -92,7 +91,7 @@ def import_failure(rvalue: Any = None):
     :param rvalue: the value to return
     """
     def decorator(func):
-        @wraps(func)
+        @functools.wraps(func)
         def wrapper(*args, **kwargs):
             try:
                 return func(*args, **kwargs)
@@ -567,7 +566,7 @@ class State(metaclass=StateMeta):
         self._state_proxy: StateProxy = StateProxy(raw_state)
         self.ingest(raw_state)
 
-        # Cette étape enregistre les propriétés associés à l'instance
+        # This step saves the properties associated with the instance
         for attribute in calculated_properties_per_state_type.get(self.__class__, []):
             getattr(self, attribute)
 
@@ -704,7 +703,7 @@ class State(metaclass=StateMeta):
         for p in path_list:
             state_proxy = self._state_proxy
             path_parts = p.split(".")
-            final_handler = partial(handler, self)
+            final_handler = functools.partial(handler, self)
             for i, path_part in enumerate(path_parts):
                 if i == len(path_parts) - 1:
                     local_mutation = MutationSubscription(path_parts[-1], final_handler)
--- a/src/writer/core.py
+++ b/src/writer/core.py
@@ -18,6 +18,7 @@ import traceback
 import types
 import urllib.request
 from abc import ABCMeta
+from contextvars import ContextVar
 from multiprocessing.process import BaseProcess
 from types import ModuleType
 from typing import (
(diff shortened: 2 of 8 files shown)


src/writer/__init__.py
```diff
@@ -93,6 +96,11 @@ def init_state(raw_state: Dict[str, Any], schema: Optional[Type[S]] = None) -> U
         raise ValueError("Root schema must inherit from WriterState")
 
     _initial_state: S = new_initial_state(concrete_schema, raw_state)
+
```
FabienArcellier, line 99: to remove, it's not required anymore

src/writer/core.py
```diff
@@ -419,6 +442,7 @@ class StateMeta(type):
     def __new__(cls, name, bases, attrs):
         klass = super().__new__(cls, name, bases, attrs)
         cls.bind_annotations_to_state_proxy(klass)
+        cls.bind_calculated_properties(klass)
```
FabienArcellier, line 445: to remove, the implementation is different

src/writer/core.py
```diff
@@ -541,16 +561,15 @@ def bind_annotations_to_state_proxy(cls, klass):
...
 class State(metaclass=StateMeta):
-    """
-    `State` represents a state of the application.
-    """
-
     def __init__(self, raw_state: Dict[str, Any] = {}):
         self._state_proxy: StateProxy = StateProxy(raw_state)
         self.ingest(raw_state)
 
+        # Cette étape enregistre les propriétés associés à l'instance
```
FabienArcellier, line 569: translate in english

src/writer/core.py
```diff
@@ -659,6 +678,48 @@ def _set_state_item(self, key: str, value: Any):
...
+        """
+        if isinstance(path, str):
+            path_list = [path]
+        else:
+            path_list = path
+
+        for p in path_list:
+            state_proxy = self._state_proxy
+            path_parts = p.split(".")
+            final_handler = functools.partial(handler, self)
```
FabienArcellier, line 706: This way of doing things is too light.

I will have to check the arguments exposed on the handler here. To be consistent, it would be necessary to inject a payload with the previous value and the current value, add the ui and add a context which indicates the mutation causing the change.

FabienArcellier:
- [ ] check that the signature of a property is valid (a single self argument), otherwise display a warning and ignore it
- [ ] check the signature of mutation event (support of state, payload, context and ui)

src/writer/core.py
```diff
@@ -659,6 +678,48 @@ def _set_state_item(self, key: str, value: Any):
                 self._state_proxy[key] = value
 
 
+    def subscribe_mutation(self, path: Union[str, List[str]], handler: Callable[['State'], None], initial_triggered: bool = False) -> None:
```
madeindjs, line 681: If you want to support array, the `path` should be able to contain number since string can't be index

```python
>>> obj = {"a": [1,2,3]}
>>> obj["a"]["0"]
  File "<stdin>", line 1, in <module>
TypeError: list indices must be integers or slices, not str
```

src/writer/core.py
```diff
@@ -659,6 +678,48 @@ def _set_state_item(self, key: str, value: Any):
...
+        :param func: handler to call when the path is mutated
+        """
+        if isinstance(path, str):
+            path_list = [path]
+        else:
+            path_list = path
+
+        for p in path_list:
+            state_proxy = self._state_proxy
+            path_parts = p.split(".")
```
madeindjs, line 705: This doesn't handle key containing `.` .

Also, it doesn't work with Array index because the `path_parts` is an array of string (see my comment below)

IMO

- it should handle a path like this one `foo["bar.baz"][0]` we might need a library to handle it
- Or we should only accept array of string/number like `["foo", "bar.baz", 0]`
- Or we considerate it's OK, but it might need a comment in the doc saying this limitation

FabienArcellier: **usage of `.` into a state key trigger errors**. That's a good point, we don't manage properly the case where a state contains `.` in the key. Currently it raise many issue. 

I would advocate to manage the character `.`  as `reserved keyword` for State key. We should ignore this value and show warning when the application is loaded.

image

**mutation of State are not monitored on list** : the `State` structure on the backend side supervises mutations exclusively on dictionaries, not on lists. This is an intrinsic limitation of WF.

A developer has to reassign a list to trigger the mutation. This point is covered in the [Mutation detection](https://dev.writer.com/framework/event-handlers#mutation-detection) in doc. It is therefore not possible to subscribe to a mutation on an element of a list.

madeindjs: Clear, thanks for the explanations.

About array, is it something we want to support ? Use case I see is to get the sum of elements in an array to get the sum. Something like this

```python
class MyState(wf.WriterState):
  arr: List[int]
  total: int

def cumulative_sum(state: MyState):
  state.total = sum(state.arr)

initial_state = ss.init_state({
    "arr": [1,2,3,4],
    "total": 0
}, schema=MyState)

initial_state.subscribe_mutation('counter', cumulative_sum)

initial_state['counter'][0] = 9
```

FabienArcellier: I thin this question has already been debated in the past. I don't find trace about that. I think we decided that for the moment we wanted to avoid this pattern. Re-encoding a list as an observable object would have an overhead that we want to avoid. 

In Python, lists are more often reconstructed rather than modified.

However, this was before the existence of typed states. Typed states allow you to specify how to behave with respect to a collection. I wonder on which usecase we would benefit from implementing this pattern.

ramedina86 (approved):
That was a tough read but looks good to me. Should I merge?

FabienArcellier:
> That was a tough read but looks good to me. Should I merge?

You can

PR #497, opened 2024-07-25
fix(ui): fix `CoreDataFrame` download for big file. WF-39

madeindjs:
The issue was too many parameters was given to `String.fromCharCode()` and make and error like "Maximum call stack size exceeded".

In fact, there is no need to try to convert the CSV to base64. So I simply used `Blob` and then `URL.createObjectURL` instead which should be more efficient.

Note, it also improve the mime/type declared, which make the OS open the file with a better software (Libreoffice instead of Text Editor for me)


FabienArcellier (commented):
I test with a dataframe of 1 million lines with success.

PR #505, opened 2024-07-30
feat(ui): implement new Slider. WF-37

madeindjs:
WIP

https://github.com/user-attachments/assets/1fe22722-bf4a-497e-bc48-3f56c4882b41


src/ui/src/core_components/base/BaseInputRange.vue
```diff
@@ -0,0 +1,212 @@
...
+	function onMouseMove(event: MouseEvent) {
+		const progress =
+			(event.x - sliderBoundingRect.left) /
+			(sliderBoundingRect.right - sliderBoundingRect.left);
+
+		if (progress > 1 || progress < 0) return;
+
+		const value = Math.round(
+			(props.max - props.min) * progress + props.min,
+		);
```
madeindjs, line 102: We might not want to round since `min/max` can be floating. We might want to round using `props.step`

src/ui/src/core_components/base/BaseInputRange.vue
```diff
@@ -0,0 +1,212 @@
...
+// clamp(0px, calc(62% - 9px), calc(100% - 18px))
+const thumbLeft = computed(
+	() => `clamp(
+		0px,
+		calc(${progress.value}% - ${thumbRadius}px),
+		calc(100% - ${thumbRadius * 2}px)
+	)`,
+);
+
+const popoverLeft = computed(() => `calc(${thumbLeft.value} - 3px)`);
```
madeindjs, line 78: Find a way to avoid this magic number

src/ui/src/core_components/base/BaseInputRange.vue
```diff
@@ -0,0 +1,212 @@
...
+
+	if (popoverTimeout) {
+		clearTimeout(popoverTimeout);
+		popoverTimeout = null;
+	}
+
+	popoverTimeout = setTimeout(() => {
+		isPopoverDisplayed.value = false;
+		popoverTimeout = null;
+	}, 1_000);
```
madeindjs, line 66: Introduce a debouncer method

```ts
function debounce<T extends (...args: unknown[]) => void>(callback: T, ms: number): (...args: Parameters<T>) => void {
  let timer;

  return (...args) => {
    if (timer) {
      clearTimeout(timer);
      timer = null;
    }

    setTimeout(() => {
      callback(...args);
      timer = null;
    }, ms);
  };
}

const hidePopover = debounce(() => (isPopoverDisplayed.value = false), 1_000);
```

FabienArcellier:
The interaction is smoother with this version. 

I realized that the graphic customization is incomplete.

Pasted image 20240809103041

Is this a problem that can be solved in this PR? Here are suggested fields.

|              |                                 |
| ------------ | ------------------------------- |
| accent       | apply on bar<br>apply on cursor |
| primary text | apply on title                  |
| popover text | apply on popover text           |
| popover      | apply on background popover     |

ramedina86:
I agree with fields suggested by @FabienArcellier but...

We need to transform the accent, otherwise the ball will have the same color as the bar, which isn't a great look. You can use the `colorTransformations`, please take a look at `CoreSwitchInput` for example where the color transformations are used.

madeindjs:
I found that we already have `--softenedAccentColor` which does the job

https://github.com/writer/writer-framework/blob/b35cee45527e8962f27d0f83e654d4f040b53d21/src/ui/src/renderer/colorTransformations.css#L11-L13

And I also added `popoverColor` & `popoverBackgroundColor` ([commit](https://github.com/writer/writer-framework/compare/754df6626cb0c2efdcc188aacd4133be6f479477..ce4082bc419a1ad1cb45fb78dfd1da0231232b35))

https://github.com/user-attachments/assets/3c30ddd8-572f-416f-9f19-333d19cf0a5f

Now we are close to the perfection :grin:

PR #511, opened 2024-08-08
feat(ui): add icon to `WdsTextInput`. WF-36

madeindjs:
Simply add `leftIcon` props to `WdsTextInput` and add a search icon to the dataframe component.

https://github.com/user-attachments/assets/1145b0f6-4d0c-4319-acea-ae932fc6a38f

### DIFF
--- a/src/ui/src/wds/WdsTextInput.vue
+++ b/src/ui/src/wds/WdsTextInput.vue
@@ -1,13 +1,35 @@
 <template>
-	<input v-model="model" type="text" />
+	<div
+		v-if="leftIcon"
+		class="WdsTextInput WdsTextInput--leftIcon"
+		@click="input.focus()"
+	>
+		<i class="material-symbols-outlined">{{ leftIcon }}</i>
+		<input
+			ref="input"
+			v-model="model"
+			type="text"
+			:placeholder="placeholder"
+		/>
+	</div>
+	<input v-else v-model="model" type="text" class="WdsTextInput" />
 </template>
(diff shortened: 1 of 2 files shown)


src/ui/src/wds/WdsTextInput.vue
```diff
@@ -19,8 +41,28 @@ input {
...
 }
 
-input:focus {
+.WdsTextInput:focus,
+.WdsTextInput--leftIcon:focus-within {
 	border: 1px solid var(--softenedAccentColor);
 	box-shadow: 0px 0px 0px 3px rgba(81, 31, 255, 0.05);
 }
+
+.WdsTextInput--leftIcon {
```
ramedina86, line 50: Hey I guess nothing wrong with it but just wanted to know, why do you do 

`.WdsTextInput--leftIcon`

instead of

`.WdsTextInput.leftIcon`

It'd more compact since

`class="WdsTextInput WdsTextInput--leftIcon"`

becomes

`class="WdsTextInput leftIcon"`

Since the styles are scoped there's no risk of another `leftIcon` bothering you.

madeindjs: It's just a habit of following [BEM convention](https://getbem.com/naming/). It helps to improve the organization of the CSS. So the `*--leftIcon` acts as a "modifier". A modifier means to change appearance, behavior or state of an element.

But I agree that here, it's not necessary since the style is scoped

Issue #354, opened 2024-03-27
checkbox for inversion of visibility value


SergeyPriem:
It seems to be useful to have checkbox near visibility selector which  appears in custom mode to invert value of visibility state. It can reduce quantity of states in application. So we can control visibilities of mutually exclusive elements with one state.
![invert checkbox](https://github.com/streamsync-cloud/streamsync/assets/97326127/18efdeee-f779-42ea-8410-aaea79eb20c6)
 code

PR #512, opened 2024-08-10
feat: checkbox for inversion of visibility value

FabienArcellier:
implement #354 

This PR implement the reverse option that allow a developer to reverse the visibility logic. It also implements binding persistence when the user returns to `Yes`

Peek 2024-08-10 07-10

### DIFF
--- a/src/ui/src/builder/BuilderSettingsVisibility.vue
+++ b/src/ui/src/builder/BuilderSettingsVisibility.vue
@@ -11,7 +11,7 @@
 					:class="{
 						active:
 							typeof component.visible == 'undefined' ||
-							component.visible === true,
+							component.visible.expression === true,
 					}"
 					@click="() => setVisibleValue(component.id, true)"
 				>
@@ -19,26 +19,39 @@
 				</div>
 				<div
 					class="chip"
-					:class="{ active: component.visible === false }"
+					:class="{ active: component.visible?.expression === false }"
 					@click="() => setVisibleValue(component.id, false)"
 				>
 					No
 				</div>
 				<div
 					class="chip"
-					:class="{ active: typeof component.visible === 'string' }"
-					@click="() => setVisibleValue(component.id, '')"
+					:class="{
+						active: component.visible?.expression === 'custom',
+					}"
+					@click="
+						() =>
+							setVisibleValue(
+								component.id,
+								'custom',
+								component.visible?.binding,
+								component.visible?.reversed,
+							)
+					"
 				>
 					Custom
 				</div>
 			</div>
 			<div
-				v-if="typeof component.visible === 'string'"
+				v-if="
+					typeof component.visible != 'undefined' &&
+					component.visible.expression === 'custom'
+				"
 				class="fieldWrapper"
 			>
 				<span class="name">Visibility value</span>
 				<BuilderTemplateInput
-					:value="component.visible"
+					:value="component.visible.binding"
 					type="state"
 					class="content"
 					placeholder="my_visibility_state_value"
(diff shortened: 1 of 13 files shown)


src/ui/src/builder/BuilderSettingsVisibility.vue
```diff
@@ -47,9 +47,24 @@
...
+						type="checkbox"
+						:checked="component.visibleReversed"
+						@input="
+							(ev: Event) =>
+								setVisibleValue(
+									component.id,
+									component.visible,
+									(ev.target as HTMLInputElement).checked,
+								)
+						"
```
madeindjs, line 65: There is a shorter syntax
```suggestion
						@input="setVisibleValue(component.id, component.visible, $event.target.checked)"
```

src/writer/core_ui.py
```diff
@@ -44,6 +44,7 @@ class Component(BaseModel):
     parentId: Optional[str] = None
     handlers: Optional[Dict[str, str]] = None
     visible: Optional[Union[bool, str]] = None
+    visibleReversed: Optional[Union[bool]] = None
```
madeindjs, line 47: It's just a bool, no ?
```suggestion
    visibleReversed: Optional[bool] = None
```

madeindjs (approved):

ramedina86 (changes requested):
Let's do this without modifying the root of the component.

Avoid having `visible` and `visibleReversed` at top level. They're closely related and should form one entity. Reversed it's also not relevant for `undefined`, `true` or `false`.

We can do 

```py
visible: {
  "reversed": true,
  "expr": "active"
}
```

So the TS type would be something like... 

```js
boolean | { "reversed": boolean, "expr": string }
```

FabienArcellier:
> ```python
> visible: {
>   "reversed": true,
>   "expr": "active"
> }
> ```

I was hesitant to make it a single entity. I gave it up. Certainly the 2 attributes live together but they do not have any particular management rule like `content` could have. These are **component parameters**. 

I would therefore prefer to **keep them at the same level** as the other parameters of a component because they have the same level of importance. The interest in nested unnecessarily complicates the data structure.

I would rather rename them:

```
{
	"visibleExpr": ""
	"visibleReverse"
}
```
Would that be enough?

ramedina86:
I'd prefer not, since `visibleReversed` is only relevant when there's an expression, not when there's `undefined`, `true` or `false`.

Another option can be to include support for `!` in the expression evaluator (or hack it and store it inside the expression at the start). I believe the state elements are `\w` regex so you could easily do that.

ramedina86:
Makes sense and this looks good, happy to move forward with it.

Let's make sure to make it backwards compatible (I'm guessing should be an easy fix via `auditAndFix`)

```py
{
	visible: {
		expression: True | False | "custom"
		binding: str
		reversed: bool
	}
}
```

src/ui/src/builder/BuilderSettingsVisibility.vue
FabienArcellier: My IDE raises a typing error with this syntax. 

image

src/writer/audit_and_fix.py
```diff
@@ -0,0 +1,46 @@
+"""
+This module allows you to migrate obsolete data structures to newer data structures.
+For example, when loading the ui.
+"""
+
+
+def fix_components(components: dict) -> dict:
```
FabienArcellier, line 7: @ramedina86 Since the implementation of pydantic, the structure of the ui.json file has been evaluated on the backend side. I had to implement the fix processing when loading the file before it was integrated into pydantic.

ramedina86: That's a good point...

FabienArcellier:
> Hey, what's the status of this?

I think it's ready to merge

PR #514, opened 2024-08-13
feat(ui): harmonize CSS cursors. WF-35

madeindjs:
Setup `cursor: pointer` for differents core components and also for differents parts of the application.


ramedina86 (commented):
It's the small things... well done

PR #516, opened 2024-08-13
feat(ui): add filter to `BuilderSidebarToolbar`. WF-42

madeindjs:
I introduced a component named `BuilderSidebarTitleSearch` which extract the existing search logic of `BuilderSidebarTree` and use it for `BuilderSidebarToolbar`.

I also took the opportunity to improve the accessibility since we use `<i>` elements as button, but they're not focusable and clickable with the keyboard.

And finally, I introduced an E2E test which cover the new feature.

https://github.com/user-attachments/assets/21771c20-02db-4a43-a27a-be11632fb209


ramedina86:
Looks great! Thanks for adding tests

PR #519, opened 2024-08-15
fix: npm run custom.dev run dev server instead of custom.dev

### LINKED ISSUE #518: npm run custom.dev run dev server instead of custom.dev
When you run the `npm run custom.dev` command, it launches the `npm run -w writer-ui dev` server insted of `npm run -w writer-ui custom.dev`.

Custom components are not displayed

FabienArcellier:
fix #518

I set the port to match the one specified in the documentation.

image
---


ramedina86 (commented):
Makes total sense, thanks

PR #485, opened 2024-07-04
feat: implement editable dataframe to manage dataframe editor component

FabienArcellier:
implement https://github.com/FabienArcellier/writer-framework/issues/65

Peek 2024-07-06 08-36

PR #529, opened 2024-08-20
feat: Implement Data frame editor.WF-66

madeindjs:
Implement behavior to edit a dataframe started in https://github.com/writer/writer-framework/pull/485 (https://github.com/FabienArcellier/writer-framework/issues/63)

https://github.com/user-attachments/assets/24cc266b-f9de-4543-8085-c407237a8dcb

For now, it only allows editing existing cells, but it'll be possible to add/delete some rows in the future.

Technically:

1. I introduced a new component `CoreDataframeCell` which re-render `CoreDataframeCellText[Text|Number|Boolean|Unknown]`
2. those component emits on changes, this event is cached by `CoreDataframe`
3. `CoreDataframe` handle this event and call a new hook `useDataFrameValueBroker` which handle the update and take care of updating the `arquero` table


src/ui/src/core_components/content/CoreDataframe/CoreDataframeCellUnknown.vue
madeindjs, line 1: add elipsis overflow

```css
text-overflow: ellipsis;
overflow: hidden;
white-space: nowrap;
display: inline-block;
```

FabienArcellier:
How do I get to this page while navigating in the hello application?
image

2 proposals :
* move the editable dataframe into tabulation instead of it's own page
* create a section with a link into tab to keep it in a separate page

For creating a section with link into tab, I would recommand those changes :
image
image

FabienArcellier:
I have been looking for adding a line for few seconds. I am not sure what the good option is.

An option is to have something more consistent with the action menu. Maybe the same format than dropdown on the bottom left.

image

FabienArcellier:
Boolean should use the same cursor than other type to show you can edit

image
image

FabienArcellier:
(not related to this PR directly) The csv export add a column `__internal_arquero_id`. I think this column should not be here.

image

FabienArcellier:
There is a bug after using the search control. The edit will change all the row. I haven't investigate the cause yet.

Peek 2024-09-14 08-16

apps/hello/main.py
```diff
@@ -126,11 +138,34 @@ def _update_scatter_chart(state):
...
+    state['editable_df'].record_update(payload)
+
+# Subscribe this event handler to the `wf-dataframe-action` event
+def on_editable_df_record_action(state, payload):
+	record_index = payload['record_index']
+	if payload['action'] == 'remove':
+		state['editable_df'].record_remove(payload)
+	if payload['action'] == 'open':
+		state['editable_df_open_text'] = str(state['editable_df'].record(record_index)['text'])
+
```
FabienArcellier, line 161: Here is a suggestion for the `test` action. 

image

```suggestion
# Subscribe this event handler to the `wf-dataframe-action` event
def on_editable_df_record_action(state, payload):
    record_index = payload['record_index']
    if payload['action'] == 'remove':
        state['editable_df'].record_remove(payload)
    if payload['action'] == 'open':
        state['editable_df_open_text'] = str(state['editable_df'].record(record_index)['text'])
    if payload['action'] == 'test':
        state.add_notification("info", "Test", "select : " + str(state['editable_df'].record(record_index)))
```

apps/hello/main.py
```diff
@@ -126,11 +138,34 @@ def _update_scatter_chart(state):
...
+	state['editable_df'].record_add(payload)
+
+
+# Subscribe this event handler to the `wf-dataframe-update` event
+def on_editable_df_record_change(state, payload):
+    payload['record']['object']['updatedAt'] = str(datetime.datetime.now())
+    state['editable_df'].record_update(payload)
+
+# Subscribe this event handler to the `wf-dataframe-action` event
+def on_editable_df_record_action(state, payload):
```
FabienArcellier, line 155: This method was using `tab` instead of `space`.

src/ui/src/components/core/base/BaseDropdown.vue
```diff
@@ -0,0 +1,133 @@
...
+function onItemClick(key: string) {
+	emits("selected", key);
+	closePopover();
+}
+
+onMounted(() => window.addEventListener("scroll", closePopover));
+onUnmounted(() => window.removeEventListener("scroll", closePopover));
+</script>
+
+<template>
```
FabienArcellier, line 47: In all our component, the `template` go first, then we have `script setup`, then `script` then style. It's different here. I would prefer to stick to the existing convention if we can.

src/ui/src/components/core/content/CoreDataframe/CoreDataframeRow.vue
```diff
@@ -0,0 +1,76 @@
...
+	actions: {
+		type: Object as PropType<Record<string, string>>,
+		required: true,
+	},
+});
+
+defineEmits({
+	change: (columnName: string, id: number, value: unknown) =>
+		typeof columnName === "string" &&
+		typeof id === "string" &&
```
FabienArcellier, line 23: are you sure about this one ? The type id is declared as number and the expected type is string ?

Is change : id, same than action: index ? If they are, you should use the same variable name

madeindjs: You're right, it's a mistake, it should be number

apps/hello/main.py
madeindjs: Good idea, implemented

ramedina86:
> For now, it only allows editing existing cells, but it'll be possible to add/delete some rows in the future.

I believe this is inaccurate?

ramedina86:
Can you add a better looking plus (+) button? With material symbols and using the WDS button

madeindjs:
@ramedina86 , do you mean to restore the buttons a bit like I did before ? If yes, using `<WdsButton size="small" />` give me this result, WDYT ?

| secondary | tertiarry |
|---|---|
| image | image |

ramedina86:
That looks good to me, why was it changed? What's the downside?

madeindjs:
> That looks good to me, why was it changed? What's the downside?

I implemented it @ramedina86

PR #534, opened 2024-08-26
fix: getting the BaseControlBar import to AnnotatedText component

polymorpheuz:


ramedina86:
That was an easy fix 😅

PR #536, opened 2024-08-26
fix: custom stylesheet and js are not loaded after OIDC authentication

### LINKED ISSUE #535: custom stylesheet and js are not loaded after OIDC authentication
Custom style sheets and js files are not loaded after a user has authenticated through an OIDC provider.

**Root cause :**

* user load the application with custom style and custom js file
* user is redirect to oidc provider portal to be authentified
* user come back after authentication. session is reused but custom style and custom js file are not send.

FabienArcellier:
This resets the mails when the session is retrieved again in the initialization process. `/static` assets can be retrieved without authentication.

fix #535

### DIFF
--- a/src/writer/app_runner.py
+++ b/src/writer/app_runner.py
@@ -132,7 +132,7 @@ class AppProcess(multiprocessing.Process):
 
         import writer
 
-        session = writer.session_manager.get_session(payload.proposedSessionId, restore_intial_mails=True)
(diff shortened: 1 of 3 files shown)


ramedina86:
Looks good but can you fix `restore_intial_mails`?

Should be `restore_initial_mail` (typo in `initial` and using singular is more in line with the rest of the code)

ramedina86:
Awesome, thanks 🙏🏻

Issue #413, opened 2024-05-08
pick up the URL variables at the start of the app


thondeboer:
I have an app that I want to be able to use URL variables, and I can figure out how to setup a handler when the values CHANGE, but how do I get the INITIAL settings of the URL values?
There is no equivalent state.get_route_vars() it seems so how do I get those values at startup?

FabienArcellier:
Hi,

Did you try to use the session object with the http header `referer` ? You have access to the url at the origin of the call. You should be able to parse it with urlparse lib.

```
def handle_app_enter(state, session):
	print(session['headers'].get('referer'))
```

An event is missing to capture a user's entry into the application at the root level. You must necessarily use the one on the page `wf-page-open`.

We can imagine a `wf-app-open` event.

thondeboer:
Thanks. I got quite confused with the rebranding. Did I miss the announcement?Regards,ThonOn Jun 5, 2024, at 23:05, Fabien Arcellier ***@***.***> wrote:﻿
Hi,
Did you try to use the session object with the http header referer ? You have access to the url at the origin of the call. You should be able to parse it with urlparse lib.
def handle_app_enter(state, session):
	print(session['headers'].get('referer'))

An event is missing to capture a user's entry into the application at the root level. You must necessarily use the one on the page wf-page-open.
We can imagine a wf-app-open event.

atzimot:
any news on this issue ?

FabienArcellier:
We haven't planned the work on that yet.

ramedina86:
We've revisited this and we've decided to move forward with a new event in _Root_ that'll be able to get URL information.

FabienArcellier:
it's available in [release `0.8.1`](https://pypi.org/project/writer/)

PR #538, opened 2024-08-26
feat: pick up the URL variables at the start of the app - WF-55

FabienArcellier:
implement #413 

Peek 2024-08-26 17-35

### TODO

* modify documentation on section https://dev.writer.com/framework/page-routes#retrieving-the-values

### DIFF
--- a/src/ui/src/components/core/root/CoreRoot.vue
+++ b/src/ui/src/components/core/root/CoreRoot.vue
@@ -17,15 +17,14 @@ import { useEvaluator } from "@/renderer/useEvaluator";
 const ssHashChangeStub = `
 def handle_hashchange(state, payload):
 	# The payload is a dictionary with the page key and all the route variables in the URL hash.
-	# For example, if the current URL is
-	# http://localhost:3000/#main/animal=duck&colour=yellow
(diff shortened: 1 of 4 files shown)


src/ui/src/components/core/root/CoreRoot.vue
```diff
@@ -107,12 +135,12 @@ watch(activePageId, (newPageId) => {
 	changePageInHash(pageKey);
 });
 
-type ParsedHash = {
+type UrlParams = {
```
ramedina86, line 138: I understand `ParsedHash` isn't an ideal name, but `UrlParams' makes me think of querystring, which this isn't.

@raaymax suggested `UrlHashParams` can we change?

src/ui/src/components/core/root/CoreRoot.vue
```diff
@@ -58,6 +74,10 @@ export default {
 			...sharedStyleFields,
 		},
 		events: {
+			"wf-app-open": {
+				desc: "Captures the first application load, capture the page key and route vars.",
```
ramedina86, line 78:
```suggestion
				desc: "Captures the first application load, including the page key and route vars.",
```

PR #539, opened 2024-08-26
feat: Abstract components

ramedina86:
As discussed, Abstract Templates allow users to specify a combination of `baseType` and `writer` (`WriterComponentDefintion`), to create a new component type that merges an existing Vue template with a new, backend-defined component definition.

The goal is to use this for Workflows, where "workflow node" will be a generic template and, for example, "Set state" will be defined from the backend, using "workflow node" as the base type.

### DIFF
--- a/src/ui/src/core/templateMap.ts
+++ b/src/ui/src/core/templateMap.ts
@@ -126,8 +126,6 @@ const templateMap = {
 
 const abstractTemplateMap: Record<string, AbstractTemplate> = {};
 
-templateMap["sectiondiff"] = CoreSection;
-
 if (WRITER_LIVE_CCT === "yes") {
 	/*
 	Assigns the components in custom_components to the template map,
--- a/src/ui/src/core/templateMap.ts
+++ b/src/ui/src/core/templateMap.ts
@@ -63,7 +63,7 @@ import CoreRoot from "../core_components/root/CoreRoot.vue";
 
 import CoreMapbox from "../core_components/embed/CoreMapbox.vue";
 
-import { WriterComponentDefinition } from "../writerTypes";
+import { AbstractTemplate, WriterComponentDefinition } from "../writerTypes";
 import { h } from "vue";
 
 const templateMap = {
@@ -124,6 +124,10 @@ const templateMap = {
 	jsonviewer: CoreJsonViewer,
 };
 
+const abstractTemplateMap: Record<string, AbstractTemplate> = {};
+
+templateMap["sectiondiff"] = CoreSection;
+
 if (WRITER_LIVE_CCT === "yes") {
 	/*
 	Assigns the components in custom_components to the template map,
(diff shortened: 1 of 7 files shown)


src/ui/src/core/templateMap.ts
```diff
@@ -163,8 +167,29 @@ function fallbackTemplate(type: string) {
...
+function getMergedAbstractTemplate(type: string) {
+	const template = abstractTemplateMap[type];
+	if (!template) return;
+	const baseType = template.baseType;
+	return {
+		...templateMap[baseType],
+		writer: {
+			...templateMap[baseType].writer,
+			...abstractTemplateMap[type].writer,
+			fields: {
```
ramedina86, line 177: Any better way to merge deep without causing trouble appreciated

FabienArcellier:
**there are 2 types of template, components and steps**. I would advocate that to encode this difference in `AbstractTemplate` . We plan to use `component` and `step` in two different menu for example.

```typescript
export type AbstractTemplate = {
	type: string; # literal (component, step)
	name: string; # maybe to rename into name
	writer: WriterComponentDefinition;
};
```

I would advocate transforming templateMap into a list and adding a function to retrieve a template. Given the size of the table, the impact should be minimal. It would also simplify the merge code.

```
templateMap = [
	CoreRoot,
	CorePage
]
```

Here is the functions to write / rewrite.

```typescript
export function getTemplate(type, name)
export function isTemplate(type, name)
export function registerComponentTemplate(type, name, vue_component)
```

src/ui/src/core/templateMap.ts
FabienArcellier: This pattern is usually call deep merge in javascript. I think there is simple way to do that with stdlib. You need to either build your own function deep merge function, either use library as lodash.

https://www.geeksforgeeks.org/how-to-deep-merge-two-objects-in-typescript/

src/writer/serve.py
```diff
@@ -150,7 +150,8 @@ def _get_run_starter_pack(payload: InitSessionResponsePayload):
             components=payload.components,
             userFunctions=payload.userFunctions,
             extensionPaths=cached_extension_paths,
-            featureFlags=payload.featureFlags
+            featureFlags=payload.featureFlags,
+            abstractTemplates=abstract.templates
```
FabienArcellier, line 154: **I like the idea of ​​having a single mechanism for reporting different forms of definitions interpretable by the frontend**.

src/writer/abstract.py
```diff
@@ -0,0 +1,16 @@
...
+
+They're used to register new component templates, based on existing frontend templates and a backend-provided component definition. 
+
+"""
+
+from typing import Dict
+
+from writer.ss_types import AbstractTemplate
+
+templates:Dict[str, AbstractTemplate]  = {}
```
FabienArcellier, line 13: I would advocate to use a list here (see my other message about the design of templateMap).

FabienArcellier (approved):

ramedina86:
It'd be very costly to change the `type` of components and I don't see a strong enough justification. The `type` is used all across Framework, in places like `ComponentProxy`, for E2E tests, etc.

`name` is already being used for the public-facing name of the component such as _Column Container_.

I'm planning on defining a new attribute on the definition called `toolkit` which you'll be able to use to define `default` or `workflows`.

PR #544, opened 2024-08-28
fix: custom stylesheet and js are not loaded after OIDC authentication

### LINKED ISSUE #535: custom stylesheet and js are not loaded after OIDC authentication

FabienArcellier:
fix #535 

The problem was not fully solved, and worse, has broke the OIDC authentication mechanism. The access to /static url is blocked after user authentication. I fail to diagnose this issue because I was loading those assets through browser cache in local during my test.

image

The problem is visible on [dev review environment](https://streamsync-dev.osc-fr1.scalingo.io/auth_oidc_auth0/). It is fixed on this [PR review environment](https://streamsync-dev-pr544.osc-fr1.scalingo.io/auth_oidc_auth0/).

---
* [x] load writer framework assets
* [x] load user assets into statics
* [x] load user extensions

### DIFF
--- a/src/writer/auth.py
+++ b/src/writer/auth.py
@@ -176,6 +176,7 @@ class Oidc(Auth):
     scope: str = "openid email profile"
     callback_authorize: str = "authorize"
     url_userinfo: Optional[str] = None
+    app_static_public: bool = False
 
     authlib: OAuth2Session = None
     callback_func: Optional[Callable[[Request, str, dict], None]] = None # Callback to validate user authentication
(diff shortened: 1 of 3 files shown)


src/writer/auth.py
```diff
@@ -191,14 +191,15 @@ def register(self,
         redirect_url = urljoin(self.host_url, self.callback_authorize)
         host_url_path = urlpath(self.host_url)
         callback_authorize_path = urljoin(host_url_path, self.callback_authorize)
-        static_assets_path = urljoin(host_url_path, "static")
+
+        auth_ignored_prefix_paths = [urljoin(host_url_path, "static"), urljoin(host_url_path, "assets")]
```
FabienArcellier, line 195: I will use the same approach than the code used in server to scan the static asset into the repo

ramedina86:
- Can't we make auth work for all paths instead of ignoring paths? There's nothing that says `static/` is public, and exposing `assets/` doesn't seem necessary either. 
- Have you tested custom components (`/extensions`)?

FabienArcellier:
I already investigate protecting static assets with WF's stateless architecture without finding a good solution. In the end, it is a common technique not to protect them, particularly to allow their management by a CDN.

The management of browser session cookies, which is supposed to be able to allow this, has been hijacked by browsers. They are persisted today. If the user refreshes the page, opens a new tab, closes and reopens the browser, the session cookie is retained.

> _Session_ cookies — cookies without a `Max-Age` or `Expires` attribute – are deleted when the current session ends. The browser defines when the "current session" ends, and **some browsers use _session restoring_ when restarting**. This can cause session cookies to last indefinitely.

Pasted image 20240830083835

Strangely, the mechanism works better for basic auth because the browser forgets the credentials when they are closed. It's also a header propagate to all the request (as cookie).

Here is the two topics where I have been stuck. Do you see a way to go further without dropping the stateless current architecture ?

**inject custom header for every query to the server**. There remains a major difficulty if we give javascript the right to access the session identifier. How to make the browser add a header to all requests, including static asset loading through an img tag. I'm stuck, it's supposed to be the role of cookies to do that. I don't know of a way to make a cookie used by a single tab. We return to the basic problem.

I look to the contextual cookie technique. It does not make it possible to manage this type of case. It is used by Google for multi-account. It only works for APIs, not for static asset loading.

FabienArcellier:
To fix the current issue, I'm hesitating between 2 strategies. Either add a blacklist in the authentication configuration, or add a whitelist in the authentication. The safest is the whitelist. It's easy to document but hard to diagnose when problem happens.

```
writer.auth.Oidc(
	protect_static_assets=[
		'/private'
		'/dataset/private_asset.csv'
	]
)
```

```
writer.auth.Oidc(
	whitelist_static_assets=[
		'/assets'
		'/dataset/asset.csv',
	]
)
```

FabienArcellier:
`/extensions` is having the same problem.

PR #545, opened 2024-08-29
feat: Component refactor and toolkit

ramedina86:
This PR introduces the concept of "toolkit", namely:
- `core` (collection of all the component templates we now until now, including _Section_, _Page_, etc) 
- `workflows` (new collection for Workflows)

It also updates the mode switcher so you can choose 4 modes:
- User Interface
- Code
- Workflows (feature flagged)
- Preview

A switch in mode determines which toolkit is shown; only the Workflows mode shows the `workflows` toolkit.

Screenshot 2024-08-29 at 12 54 20

There's also been a refactor in the `core_components` folder to allow for a neater hierarchy that can accommodate the two toolkits, `core` and `workflows`. Paths have been updated to use the "@/" syntax, since the relative syntax was getting out of hand i.e. `../../../injectionKeys` etc.


raaymax (approved):

madeindjs (approved):

src/ui/src/custom_components/BubbleMessage.vue
madeindjs, line 1: Don't you think it should be placed in `src/components/custom`

ramedina86: I didn't think it was urgent but yes, absolutely. I'll create a Jira ticket so that we don't forget.

PR #550, opened 2024-09-02
fix: custom component name should be constrained explicitely - WF-52

### LINKED ISSUE #517: custom component name should be constrained explicitely
Custom component names defined by the dev in custom_components/index.ts must contain the full original name of the component lower-case.

For example, A customized the Metric component called "customMetric", compiles but can't be dragged into the ComponentRenderer, whereas "custommetric" and "nonsensemetric" work fine. AFAIK the constraints for naming are not in the docs. video of the behavior from the invalid name is below

https://github.com/user-attachments/assets/4d9c69e6-5511-47be-90aa-d1f79447292e

FabienArcellier:
fix #517 

---
image

### DIFF
--- a/alfred/npm.py
+++ b/alfred/npm.py
@@ -5,7 +5,8 @@ import alfred
 
 @alfred.command("npm.lint", help="lint check npm packages")
 def npm_lint():
-    alfred.run("npm run ui:lint:ci")
+    alfred.run("npm run ui:lint.ci")
+    alfred.run("npm run ui:custom.check")
 
 @alfred.command("npm.e2e", help="run e2e tests")
 @alfred.option('--browser', '-b', help="run e2e tests on specified browser", default='chromium')
--- a/docs/framework/custom-components.mdx
+++ b/docs/framework/custom-components.mdx
@@ -165,13 +165,18 @@ A single or multiple templates can be specified. Take into account that they wil
 
 ## Bundling templates
 
-Execute `npm run custom.build`, this will generate the output `.js` and `.css` files into `ui/custom_components_dist`.
+Execute `npm run custom.build` into `src/ui`, this will generate the output `.js` and `.css` files into `./custom_components_dist`.
 
 ```sh
 # "build" builds the entire front-end
 # "custom.build" only builds the custom templates
 
+npm run custom.check # Optional: checks certain issues on custom components
 npm run custom.build
 ```
 
-Collect the files from `ui/custom_components_dist` and pack them in a folder such as `my_custom_bubbles`. The folder containing the generated files, e.g. `my_custom_bubbles`, can now be placed in the `extensions/` folder of any Framework project. It'll be automatically detected during server startup.
+Collect the files from `./custom_components_dist` and pack them in a folder such as `my_custom_bubbles`. The folder containing the generated files, e.g. `my_custom_bubbles`, can now be placed in the `extensions/` folder of any Framework project. It'll be automatically detected during server startup.
+
+<Tip>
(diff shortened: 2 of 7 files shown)


src/ui/src/custom_components/core.ts
```diff
@@ -0,0 +1,32 @@
...
+	return customComponents;
+}
+
+/**
+ * Audit the validity of custom components
+ *
+ * @param key
+ * @param customComponent
+ */
+export function _auditCustomComponents(key: string, customComponent: unknown) {
```
madeindjs, line 22: 1. You can remove the `_` prefix and simply not export this function
2. I think the function name could be more meaningfull
```suggestion
function checkComponentName(key: string) {
```

src/ui/src/custom_components/core.ts
```diff
@@ -0,0 +1,32 @@
...
+	// Checks that the key contains only alphanumeric characters and underscores without capital letters
+	// @see https://github.com/writer/writer-framework/issues/517
+	const regex = /^[a-z0-9_]+$/;
+	const is_valid_key = regex.test(key);
+	if (!is_valid_key) {
```
madeindjs, line 27: In JS, we prefer camlCase like `isValidKey`.
```suggestion
	const isValidKey = regex.test(key);
	if (!isValidKey) {
```

src/ui/src/custom_components/core.ts
```diff
@@ -0,0 +1,32 @@
...
+		throw Error(
+			`custom component key '${key}' is not valid. custom component should be declared with only alphanumeric lowercase and _.`,
+		);
```
madeindjs, line 30: Question, do we really want to stop the application here ? Do you think `console.warn` could be enough ?

src/ui/src/custom_components/core.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Declare custom components
+ *
+ * @param customComponents list of components in key-value form
+ */
+export function declareCustomComponents(
```
madeindjs, line 6: I think the function name could be more meaningfull. Maybe `checkCustomComponents`

FabienArcellier: In this case I want to call the function like this `declareCustomComponents`. It seems more logical when you read the calling code :

```typescript
export default declareCustomComponents({
	bubblemessage: BubbleMessage,
	bubbleMessageadvanced: BubbleMessageAdvanced,
});
```

ramedina86:
@madeindjs thanks for taking a look at this

@FabienArcellier I think we should check during load, it'll be more straightforward and reliable. Also, let's try to keep `core` as lean as we can.

We can intercept the keys here:

`importCustomComponentTemplate` in `loadExtensions.ts` (for loading via transpiled extension)
`if (WRITER_LIVE_CCT === "yes") {...` in `templateMap.ts` (for loading live while developing)

FabienArcellier:
> @madeindjs thanks for taking a look at this
> 
> @FabienArcellier I think we should check during load, it'll be more straightforward and reliable. Also, let's try to keep `core` as lean as we can.
> 
> We can intercept the keys here:
> 
> `importCustomComponentTemplate` in `loadExtensions.ts` (for loading via transpiled extension) `if (WRITER_LIVE_CCT === "yes") {...` in `templateMap.ts` (for loading live while developing)

I used this approach. I am looking for a way to raise the error during the build or using a dedicated npm command.
The problem only appears in runtime when you open the web application. It does not feel great.

FabienArcellier:
I have implemented a command `npm run custom.check` that verify this rule. It will allow to check the status without running the entire app and allow a person to implement a start of continuous integration pipeline when building custom components. I add the info in the documentation.

image

PR #556, opened 2024-09-09
chore: bump completion default model

mmikita95:
Upgrade `palmyra-x-002-instruct` to `palmyra-x-003-instruct` as per WF-65


mmikita95:
I ran explicit AI tests locally, works good:
image

Issue #225, opened 2024-02-03
Slider width is not constant if it involves decimals


crxi:
When sliders are used for real numbers with decimal places, their widths are not constant.
![image](https://github.com/streamsync-cloud/streamsync/assets/79311899/b694ec70-9efb-46c7-bdc9-3dee3d1c5f94)

This is because the value printed at the end has variable length. For example, "0.9" and "1.1" are longer than "1" so their sliders are shorter. This is problematic when we click and drag the slider from 0.9 to 1.1. When it crosses 1, the slider is  lengthen resulting in jerky movement of the caret as the slider is redrawn with different widths.

Questions
- Is there a way to make the slider and number output constant width? I don't see any option.
- Can we show intermediate marks on the slder, or at least the min and max values. It is hard for a user to know the range of possible otherwise.

FabienArcellier:
A classic way to get around the problem is to display a tooltip instead of the slider value. Here is an example taken from the material ui document which can serve as a basis for reflection.

![Peek 2024-02-04 20-41](https://github.com/streamsync-cloud/streamsync/assets/159559/9a311586-480c-438b-ac5e-5d728b45bb30)

We may add options as :

* Show start / end markers
* Show steps

Issue #561, opened 2024-09-17
support for floating numbers and large numbers on the slider


FabienArcellier:
The new slider did not support floating numbers, for example, a step of 0.1.

![Peek 2024-09-17 06-20](https://github.com/user-attachments/assets/7759e849-beb1-4365-8276-af3e43fe226b)
*a step of 0.1 is configured on this example*

The new slider was unable to display huge number as 10 000.

![image](https://github.com/user-attachments/assets/a3f9ffd7-c846-4ccf-8bb7-e774c5426b5e)

FabienArcellier:
it's available in [release `0.8.1`](https://pypi.org/project/writer/)

PR #562, opened 2024-09-17
fix: support for floating numbers and large numbers on the slider

FabienArcellier:
implement #561, improve #505

Peek 2024-09-17 06-40

The current component did not handle steps composed of decimals like 0.1 and 0.01. This PR adds support to permanently fix issue #225.

Decimal numbers can be quite large. The current implementation is of fixed size. The CSS code is modified so that the size of the popover is dynamic.

### DIFF
--- a/src/ui/src/components/core/base/BaseInputRange.vue
+++ b/src/ui/src/components/core/base/BaseInputRange.vue
@@ -4,7 +4,7 @@
 		:aria-valuemin="min"
 		:aria-valuemax="max"
 		:aria-valuenow="value"
-		:aria-valuetext="`${value} (${Math.round(progress)}%)`"
+		:aria-valuetext="`${displayValue} (${Math.round(progress)}%)`"
 		class="BaseInputRange"
 		:class="{
 			'BaseInputRange--popover-always-visible':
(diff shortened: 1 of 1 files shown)


src/ui/src/components/core/base/BaseInputRange.vue
```diff
@@ -64,6 +64,10 @@ const thumbRadius = 9;
 
 const thumb = ref<HTMLElement>();
 const slider = ref<HTMLElement>();
+const displayValue = computed(() => {
+	const precision = Math.ceil(-Math.log10(props.step));
+	return model.value.toFixed(precision);
+});
```
madeindjs, line 70: I would use two computed values, because it avoids to re-compute `precision` when value change
```suggestion
const precision = computed(() => Math.ceil(-Math.log10(props.step)));
const displayValue = computed(() => model.value.toFixed(precision.value));
```

src/ui/src/components/core/base/BaseInputRange.vue
```diff
@@ -190,6 +185,7 @@ function handleMouseDown(initialEvent: MouseEvent) {
 	top: 0;
 
 	transition: box-shadow ease-in-out 0.5s;
+	transform: translateX(-50%);
```
madeindjs, line 188: Well seen, I didn't think about using this.

madeindjs (approved):

PR #568, opened 2024-09-23
feat: the WF engine should resolves escaped variables - WF-53

FabienArcellier:
* feat: handle escape variable in exchange protocol
* feat: handle escape variable in auto-completion
* feat: state support dot variable

Peek 2024-10-07 10-10

### DIFF
--- a/src/ui/src/builder/BuilderTemplateInput.vue
+++ b/src/ui/src/builder/BuilderTemplateInput.vue
@@ -114,7 +114,9 @@ const handleComplete = (selectedText) => {
 	const full = getPath(text);
 	if (full === null) return;
 	const keyword = full.at(-1);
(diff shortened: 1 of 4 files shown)


src/ui/src/builder/BuilderTemplateInput.vue
```diff
@@ -186,7 +195,7 @@ const showAutocomplete = () => {
 const handleBlur = () => {
 	setTimeout(() => {
 		autocompleteOptions.value = [];
-	}, 100);
+	}, 300);
```
FabienArcellier, line 198: I encountered cases where the handler reaction was too fast. We may have to change this behavior in the future to have a flag to show or hide the autocompletion.

ramedina86 (approved):

PR #570, opened 2024-09-27
feat: Workflow tools

ramedina86:


ramedina86:
@madeindjs You always share some good advice :) Happy to provide more context for these features if necessary.

src/ui/src/components/workflows/WorkflowsWorkflow.vue
```diff
@@ -209,55 +235,69 @@ function handleNodeDrag(ev: DragEvent, componentId: Component["id"]) {
...
+		// Remove references when a node is deleted
+
+		const preIds = preChildren.map((c) => c.id);
+		const postIds = postChildren.map((c) => c.id);
+		const removedIds = preIds.filter((cId) => !postIds.includes(cId));
+		removedIds.forEach((removedId) => {
+			postChildren.forEach((c) => {
+				if (!c.outs || c.outs.length == 0) return;
+				c.outs = c.outs.filter((out) => out.toNodeId !== removedId);
+			});
```
madeindjs, line 292: I think you can avoid the `O(n²)` and I guess it makes sens to use `Set`. Maybe like this (not tested)
```suggestion
    const removedIds = new Set(postChildren.map((c) => c.id));
    preChildren.forEach(c => removedIds.delete(c.id));

    postChildren.forEach((c) => {
      c.outs = c.outs?.filter((out) => !removedIds.has(out.toNodeId));
    });
```
There is also the new method [`Set.difference`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference) but still not widely available ([caniuse](https://caniuse.com/?search=set%20difference))

src/ui/src/builder/useComponentActions.ts
```diff
@@ -688,6 +696,73 @@ export function useComponentActions(wf: Core, ssbm: BuilderManager) {
...
+		if (!component) return;
+		const transactionId = `edit-${componentId}-out-${out.outId}-${out.toNodeId}`;
+		ssbm.openMutationTransaction(transactionId, `Edit out`, true);
+		ssbm.registerPreMutation(component);
+
+		component.outs = [
+			...component.outs.filter(
+				(o) => !(out.outId === o.outId && out.toNodeId === o.toNodeId),
+			),
+		];
```
madeindjs, line 736: You don't have to "clone" the array again
```suggestion
		component.outs = component.outs.filter(
			(o) => !(out.outId === o.outId && out.toNodeId === o.toNodeId),
		);
```

src/ui/src/builder/useComponentActions.ts
```diff
@@ -127,6 +127,7 @@ export function useComponentActions(wf: Core, ssbm: BuilderManager) {
 		type: string,
 		parentId: Component["id"],
 		position?: number,
+		initProperties?: Partial<Component>,
```
madeindjs, line 130: Since many props are overridden

https://github.com/writer/writer-framework/blob/f17a349a11f7df87de53dc04227fbccd40b01db8/src/ui/src/builder/useComponentActions.ts#L140-L148

the type could be a bit more correct
```suggestion
		initProperties?: Partial<Omit<Component, 'id' | 'type' | 'parent' | 'content' | 'handlers' | 'position'>>,
```

src/ui/src/builder/useComponentActions.ts
```diff
@@ -160,8 +162,14 @@ export function useComponentActions(wf: Core, ssbm: BuilderManager) {
...
```
madeindjs, line 165: Same as before

src/ui/src/components/workflows/WorkflowsWorkflow.vue
```diff
@@ -209,55 +235,69 @@ function handleNodeDrag(ev: DragEvent, componentId: Component["id"]) {
...
+
+	component.x = newX;
+	component.y = newY;
+
+	setTimeout(() => {
+		// Debouncing
+		if (component.x !== newX) return;
+		if (component.y !== newY) return;
+		changeCoordinates(componentId, newX, newY);
+	}, 200);
```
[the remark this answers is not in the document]
ramedina86: Yes that was the other alternative but I wanted to avoid having a single timer `id` for all components. So I'd have needed `Record<Component["id"], ReturnType<typeof setTimeout>>` and I wanted to avoid.

However, in the example you declare `id` inside the function. If it works that'd be great, but how would that behave? Wouldn't future calls get a completely new `id` since it's scoped to the function?

src/ui/src/builder/useComponentActions.ts
ramedina86: Yes, completely unnecessary, well spotted.

src/ui/src/builder/useComponentActions.ts
ramedina86: It does make sense, yes. Thanks.

src/ui/src/components/workflows/WorkflowsWorkflow.vue
ramedina86: I implemented the suggestion but it wasn't working, I asked Palmyra to break it down.

-------

 The issue with the second alternative triggering an infinite recursion likely stems from how the `removedIds` set is being manipulated and used. Let's break down the differences between the two alternatives to understand why the second one causes an infinite loop.

### First Alternative

In the first alternative, the `removedIds` array is calculated by comparing the `preIds` and `postIds` arrays. This ensures that only the IDs that are present in `preChildren` but not in `postChildren` are considered as removed.

```javascript
const preIds = preChildren.map((c) => c.id);
const postIds = postChildren.map((c) => c.id);
const removedIds = preIds.filter((cId) => !postIds.includes(cId));
```

Then, it iterates over `postChildren` and filters out any `outs` that reference the removed IDs.

```javascript
removedIds.forEach((removedId) => {
    postChildren.forEach((c) => {
        if (!c.outs || c.outs.length == 0) return;
        c.outs = c.outs.filter((out) => out.toNodeId !== removedId);
    });
});
```

### Second Alternative

In the second alternative, the `removedIds` set is initially populated with the IDs from `postChildren`, and then the IDs from `preChildren` are removed from this set.

```javascript
const removedIds = new Set(postChildren.map((c) => c.id));
preChildren.forEach((c) => removedIds.delete(c.id));
```

However, this logic is flawed because it ends up with a set of IDs that are present in `postChildren` but not in `preChildren`, which is the opposite of what you want. You want the IDs that are present in `preChildren` but not in `postChildren`.

Additionally, the filtering logic is applied directly to `postChildren`, which might cause the watcher to trigger again if the `outs` array is modified.

```javascript
postChildren.forEach((c) => {
    c.outs = c.outs?.filter((out) => !removedIds.has(out.toNodeId));
});
```

### Why Infinite Recursion?

The infinite recursion occurs because the watcher is set to `deep: true`, which means it watches for changes in nested properties. When you modify `c.outs` within the watcher, it triggers the watcher again, leading to an infinite loop.

### Corrected Second Alternative

To fix the second alternative, you should correctly calculate the `removedIds` set and avoid modifying `postChildren` directly within the watcher. Here's a corrected version:

```javascript
watch(
    children,
    async (postChildren, preChildren) => {
        // Remove references when a node is deleted

        const preIds = new Set(preChildren.map((c) => c.id));
        const postIds = new Set(postChildren.map((c) => c.id));
        const removedIds = new Set([...preIds].filter((cId) => !postIds.has(cId)));

        postChildren.forEach((c) => {
            if (!c.outs || c.outs.length === 0) return;
            c.outs = c.outs.filter((out) => !removedIds.has(out.toNodeId));
        });

        // Refresh arrows

        await nextTick();
        refreshArrows();

        if (removedIds.size > 0) {
            wf.sendComponentUpdate();
        }
    },
    { deep: true },
);
```

In this corrected version, `removedIds` is calculated correctly, and the filtering logic is applied without causing the watcher to trigger again. This should prevent the infinite recursion.

PR #572, opened 2024-09-28
fix(ui): fix number format for the input slider

madeindjs:
When importing a slider, the initial value is `''`, which throws an error like `TypeError: "".toFixed is not a function`.

Also, we don't handle the step change correctly, which can leads to `RangeError: precision -1 out of range`

https://github.com/user-attachments/assets/abbe608b-d133-4edc-8c82-67376308f3c4

### DIFF
--- a/src/ui/src/components/core/base/BaseInputRange.vue
+++ b/src/ui/src/components/core/base/BaseInputRange.vue
@@ -65,8 +65,11 @@ const thumbRadius = 9;
 const thumb = ref<HTMLElement>();
 const slider = ref<HTMLElement>();
 
-const precision = computed(() => Math.ceil(-Math.log10(props.step)));
-const displayValue = computed(() => model.value.toFixed(precision.value));
+const precision = computed(() => String(props.step).split(".")[1]?.length ?? 0);
+const displayValue = computed(() => {
+	if (typeof model.value !== "number") return "";
+	return Number(model.value).toFixed(precision.value);
+});
 
 const progress = computed(() => {
 	if (typeof model.value !== "number") return 50;
@@ -215,6 +218,8 @@ function handleMouseDown(initialEvent: MouseEvent) {
 	justify-content: center;
 	padding: 0 4px;
 	transform: translateX(-50%);
+
+	min-width: 12px;
 }
 
 .BaseInputRange__popover::after {


src/ui/src/components/core/base/BaseInputRange.vue
```diff
@@ -65,8 +65,10 @@ const thumbRadius = 9;
...
 const slider = ref<HTMLElement>();
 
-const precision = computed(() => Math.ceil(-Math.log10(props.step)));
-const displayValue = computed(() => model.value.toFixed(precision.value));
+const precision = computed(() => String(props.step).split(".")[1]?.length ?? 0);
+const displayValue = computed(() =>
+	Number(model.value).toFixed(precision.value),
+);
 
 const progress = computed(() => {
```
FabienArcellier, line 73: regarding the behavior of model, I think we should also write this change

```
const progress = computed(() => {
	const v =
		typeof model.value !== "number" ? Number(model.value) : model.value;

	return ((v - props.min) / (props.max - props.min)) * 100;
});
```

madeindjs: It's a good point, the thing is the empty state is always a `string` because of this

https://github.com/writer/writer-framework/blob/9f4af7b7f5c5bc247bc8174a900d6c98456ef529/src/ui/src/renderer/useFormValueBroker.ts#L113

Which make things a bit weird to handle... That's why I choose to put the thumb in the middle.

In fact, I can't display `0` neither because the `min` could be greater. Thus, I think the best fix is to display nothing in this situation like this :

```ts
const displayValue = computed(() => {
	if (typeof model.value !== "number") return "";
	return Number(model.value).toFixed(precision.value);
});
```

I updated it, well spoted

FabienArcellier: I feel the experience is not great for the user.

image

and something like that ? It will render

image

```typescript
const displayValue = computed(() => {
	const v = model.value !== "" ? model.value : props.min;
	return Number(v).toFixed(precision.value);
});

const progress = computed(() => {
	const v = model.value !== "" ? model.value : props.min;
	return ((v - props.min) / (props.max - props.min)) * 100;
});
```

PR #574, opened 2024-10-01
fix(ui): make `CoreSwitchInput` reactive to state change - WF-81

madeindjs:
We used a `ref` to store the state of the input and try to synchronize with `useFormValueBroker`. But in fact, there is no reason to do it since this method already use an optimistic update.

https://github.com/writer/writer-framework/blob/9f4af7b7f5c5bc247bc8174a900d6c98456ef529/src/ui/src/renderer/useFormValueBroker.ts#L47-L52

https://github.com/user-attachments/assets/9c19a653-708d-463d-a82c-4ac7d220925c

### DIFF
--- a/src/ui/src/components/core/input/CoreSwitchInput.vue
+++ b/src/ui/src/components/core/input/CoreSwitchInput.vue
@@ -7,10 +7,10 @@
 	>
 		<div
 			class="switch"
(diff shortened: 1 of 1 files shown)


FabienArcellier:
implement #576

PR #575, opened 2024-10-01
chore(e2e): migrate `ui.json` of E2E presets

madeindjs:
Migrated the `ui.json` of E2E to avoid warning during E2E test. I just ran `writer edit tests/e2e/presets/$FILE` in each preset folder and committed.

It should avoid unnecessary computation during the E2E execution and ease the migration.


tests/e2e/index.js
```diff
@@ -126,6 +127,20 @@ class WriterProcessPool {
 		await fs.rm(`./runtime/${id}`, { recursive: true });
 	}
 
+
+}
+async function copyDir(src,dest) {
```
FabienArcellier, line 132: is there a reason to not use `fs.cp(source, destination, { recursive: true }` : https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options ?

madeindjs: Ha shit, I forgot this. I took the first answer on Stackoverflow 🤦  I'll update, well spotted

madeindjs: [Fixed](https://github.com/writer/writer-framework/compare/da05cf2ef22daf42973e8400bf12413ccef649e8..14578ca3db2259ddf7aad1c4fc7f580a18fb82aa)

PR #580, opened 2024-10-07
fix(ui): handle visibility from object prop for `CoreTab`. WF-73

madeindjs:
The issue was that the tab is randomly visible when receiving Visibility setting from the state object property.

The root cause is `ComponentProxy` try to apply `display: none`, but it collide with `CoreTab`'s `v-show`. The fix is simply to check the visibily in `CoreTab` to adapt `v-show` accordingly.


ramedina86:
Nice investigation and fix ;)

PR #584, opened 2024-10-10
chore(ui): clean Eslint error - WF-48

madeindjs:
Get rid of all Eslint warning by hand and also bump `@typescript-eslint/eslint-plugin` to avoid incompatible Typescript version warning

image


ramedina86:
Thanks for this, it was much needed

PR #597, opened 2024-10-15
feat: add deploy alias

raaymax:


ramedina86:
@raaymax any ideas on why the CI might be failing?

raaymax:
No idea, it's failing the same way also in the other PR..

raaymax:
Found this https://github.com/microsoft/playwright/issues/30368
 Seems to be related to ubuntu 24 🤔

ramedina86:
It does, so how do we fix? cc @FabienArcellier

ramedina86:
Looks like it run this time though?

src/writer/command_line.py
```diff
@@ -7,7 +7,7 @@
 import click
 
 import writer.serve
-from writer.deploy import cloud
+from writer.deploy import cloud, deploy_app
```
madeindjs, line 10: Can't we keep only `deploy` method in `deploy.py` and use an alias here ?
```suggestion
from writer.deploy import cloud, deploy as deploy_app
```

src/writer/command_line.py
```diff
@@ -73,6 +73,23 @@ def hello(port: Optional[int], host: str, enable_remote_edit):
...
+@main.command()
+@click.option('--api-key',
+    default=lambda: os.environ.get("WRITER_API_KEY", None),
+    allow_from_autoenv=True,
+    show_envvar=True,
+    envvar='WRITER_API_KEY',
+    prompt="Enter your API key",
+    hide_input=True, help="Writer API key"
+)
+@click.option('--env', '-e', multiple=True, default=[], help="Environment to deploy the app to")
```
madeindjs, line 85: Maybe `Environments` as it can be multiple

```suggestion
@click.option('--env', '-e', multiple=True, default=[], help="Environments to deploy the app to")
```

madeindjs (approved):

src/writer/command_line.py
ramedina86: Aren't these environment variables?

raaymax:
You have no idea how much time I spent in click docs searching for exactly something like this... 
Shout out to @FabienArcellier :D

src/writer/deploy.py
```diff
@@ -30,7 +30,7 @@ def cloud():
     prompt="Enter your API key",
     hide_input=True, help="Writer API key"
 )
-@click.option('--env', '-e', multiple=True, default=[], help="Environment to deploy the app to")
+@click.option('--env', '-e', multiple=True, default=[], help="Environment variable to deploy the app to")
```
ramedina86, line 33:
```suggestion
@click.option('--env', '-e', multiple=True, default=[], help="Set environment variables for the app (e.g., --env KEY=VALUE)")
```

PR #601, opened 2024-10-20
feat(ui): add allow file type for `CoreFileInput` - WF-41

madeindjs:
Just introduce a field for the HTML attribute
[`accept`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#limiting_accepted_file_types) input file.

image

### DIFF
--- a/src/ui/src/components/core/input/CoreFileInput.vue
+++ b/src/ui/src/components/core/input/CoreFileInput.vue
@@ -11,6 +11,7 @@
 					ref="fileEl"
 					type="file"
 					:multiple="allowMultipleFilesFlag"
+					:accept="allowFileTypes"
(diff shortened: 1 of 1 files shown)


ramedina86:
How do you specify "all types"? Is it just a matter of leaving it blank. If so, please specify in the `desc`.

madeindjs:
@ramedina86 , I fixed it ([compare](https://github.com/writer/writer-framework/compare/b58a5c68d03fdaaf187e893c3177597381cae292..9268825e95795650645b89a78d11e896ed5883ee)). Now it saids
> Provides hints for browsers to select the correct file types. You can specify extensions and MIME types separated by comma, or leave empty to accept any file.

PR #603, opened 2024-10-21
feat: Workflows function calling support

ramedina86:


src/ui/src/builder/BuilderFieldsTools.vue
```diff
@@ -0,0 +1,282 @@
...
+
+const toolFormInitValue = {
+	isShown: false,
+	type: "function" as "function" | "graph",
+	name: "new_tool",
+	code: initFunctionToolCode,
+	graphId: "999-999",
+};
+
+const toolForm: Ref<ToolForm> = ref(toolFormInitValue);
```
madeindjs, line 116:
```suggestion
const toolForm = ref<ToolForm>(toolFormInitValue);
```

src/ui/src/builder/BuilderFieldsTools.vue
```diff
@@ -0,0 +1,282 @@
...
+
+const tools: ComputedRef<Record<string, Tool>> = computed(() => {
+	let value = {};
+	try {
+		value = JSON.parse(component.value.content[fieldKey.value]);
+	} catch {
+		value = {};
+	}
+	return value;
+});
```
madeindjs, line 133: Could be a bit shorter
```suggestion
const tools = computed<Record<string, Tool>>(() => {
	try {
		return JSON.parse(component.value.content[fieldKey.value]);
	} catch {
		return {};
	}
});
```

src/ui/src/builder/BuilderFieldsTools.vue
```diff
@@ -0,0 +1,282 @@
...
+	let errors = [];
+	const { name } = form;
+	if (!name) {
+		errors.push("The name cannot be empty.");
+	}
+	if (Object.keys(tools.value).includes(name)) {
+		errors.push("An existing tool with the specified name already exists.");
+	}
+	return errors;
+}
```
madeindjs, line 169: Optional, but I think it's a better pattern to use `computed` here. For example.
```suggestion
const tooolFromForm = computed<Tool | null>(() => {
	const { type, code, graphId } = toolForm.value;
	if (type == "function") {
		return {
			...JSON.parse(code),
			type,
		};
	}
	if (type == "graph") {
		return {
			type,
			graph_ids: [graphId],
		};
	}
	return null;
});

const toolFormErrors = computed<string[]>(() => {
	let errors = [];
	const { name } = form;
	if (!name) {
		errors.push("The name cannot be empty.");
	}
	if (Object.keys(tools.value).includes(name)) {
		errors.push("An existing tool with the specified name already exists.");
	}
	return errors;
});
```

src/ui/src/builder/BuilderFieldsTools.vue
```diff
@@ -0,0 +1,282 @@
...
+
+const customHandlerModalCloseAction: ModalAction = {
+	desc: "Close",
+	fn: () => {
+		toolForm.value.isShown = false;
+	},
+};
+</script>
+
+<style>
```
madeindjs, line 237: can't it be scoped ?

src/ui/src/builder/BuilderFieldsTools.vue
```diff
@@ -0,0 +1,282 @@
...
+				v-for="(tool, toolName) in tools"
+				:key="toolName"
+				class="tool"
+				@click="editTool(toolName)"
+			>
+				<div class="toolName">{{ toolName }}</div>
+				<button class="delete" @click.stop="deleteTool(toolName)">
+					<i class="material-symbols-outlined">delete</i>
+				</button>
+			</div>
```
madeindjs, line 17: If it's clickable, it should be a button (with `type="button"`, otherwise it sends parent form)
```suggestion
			<button
				v-for="(tool, toolName) in tools"
				:key="toolName"
				class="tool"
				type="button"
				@click="editTool(toolName)"
			>
				<div class="toolName">{{ toolName }}</div>
				<button class="delete" @click.stop="deleteTool(toolName)">
					<i class="material-symbols-outlined">delete</i>
				</button>
			</button>
```

Otherwise, you need to declare more attributes
```suggestion
			<div
				v-for="(tool, toolName) in tools"
				:key="toolName"
				class="tool"
				tabindex="0"
				@click="editTool(toolName)"
				@keydown.enter="editTool(toolName)"
			>
				<div class="toolName">{{ toolName }}</div>
				<button class="delete" @click.stop="deleteTool(toolName)">
					<i class="material-symbols-outlined">delete</i>
				</button>
			</div>
```

src/ui/src/builder/BuilderFieldsTools.vue
```diff
@@ -0,0 +1,282 @@
...
+		</div>
+
+		<WdsButton
+			variant="builder"
```
madeindjs, line 21: This variant doesn't exist, no ?

https://github.com/writer/writer-framework/blob/aa9b61cdfe2a79e4fce7fed88962f9e407598d21/src/ui/src/wds/WdsButton.vue#L11

src/ui/src/builder/BuilderFieldsTools.vue
```diff
@@ -0,0 +1,282 @@
...
+			size="small"
+			@click="resetAndShowToolFormModal"
+		>
+			<i class="material-symbols-outlined">add</i>
+			Add tool</WdsButton
+		>
```
madeindjs, line 27: Could we create a task to have `icon` attribute in `WdsButton` ? For example, here you should have a `gap` between the icon and the test [Figma](https://www.figma.com/design/jgLDtwVwg3hReC1t4Vw20D/.WDS-Writer-Design-System?node-id=666-2324&t=DubUJE0j7EF9N4De-4)

ramedina86: It does have a gap, the `WdsButton` is actually a flex with a gap in it, so whatever you put in the slot comes with a gap.

src/ui/src/builder/BuilderFieldsTools.vue
ramedina86: > otherwise it sends parent form
I don't understand this one

ramedina86: Agreed that it needs `tabindex` and `keydown.enter` though

src/ui/src/builder/BuilderEmbeddedCodeEditor.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+	<div ref="builderEditor" class="BuilderEmbeddedCodeEditor">
+		<div ref="editorContainer" class="editorContainer"></div>
+	</div>
+</template>
+
+<script setup lang="ts">
+import * as monaco from "monaco-editor";
```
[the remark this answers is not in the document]
ramedina86: Monaco import was a mess back in the day. I'll create a task and you can look into it in detail, perform tests, etc.

src/ui/src/builder/BuilderFieldsTools.vue
ramedina86: This syntax is really nice, I didn't know it existed, thanks for bringing it up.

src/ui/src/builder/BuilderFieldsTools.vue
madeindjs: >  otherwise it sends parent form

I was referring to the `type` attribute of the `<button>`. If you don't specify the `type="button"`, it's by default `<button type="submit">` ([doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type)). This can make the button submit the parent form if there is any

```html
<form onsubmit="alert('submited')">
  <input type="submit" value="This button triggers submit form" />
  <button type="submit">This button triggers submit form</button>
  <button>This button triggers submit form</button>

  <button type="button">This button doesn't triggers submit form</button>
</form>
```

Check [this codepen](https://codepen.io/madeindjs/pen/zYgWoWo)

So it's always good to specify `type="button"` to avoid weird behavior

ramedina86: Ah I didn't know! Thought it was `button` by default

PR #611, opened 2024-10-31
chore(ui): clean Eslint error

madeindjs:


ramedina86:
I think these should all be cleaned up after the last PR I merged 😄

PR #615, opened 2024-11-05
chore(ui): move custom components - WF-63

madeindjs:
Now we have a folder to contains all components
(`src/ui/src/components/{shared,common,core}`), we can move `custom_components` into it.


ramedina86:
Have you tested creating a new custom component with this configuration? Should work; but many moving parts so I'd like to make sure

madeindjs:
> Have you tested creating a new custom component with this configuration? 

Yes, it works :raised_hands: 

image

PR #622, opened 2024-11-08
chore: serialize tool calls messages with non-empty content

mmikita95:
Enable serialization for messages that contain `tool_calls` but have non-empty `content` – in cases when model generates both `content` and `tool_calls` without separation.

### DIFF
--- a/src/writer/ai.py
+++ b/src/writer/ai.py
@@ -1817,7 +1817,8 @@ class Conversation:
         """
         Function to verify whether the message should be serializable.
 
-        :return: Boolean that indicates
+        :return: Boolean indicating if the message meets 
+        the criteria for serialization.
         """
         if message["role"] in ["system", "tool"]:
             # Prevent serialization of messages
--- a/src/writer/ai.py
+++ b/src/writer/ai.py
@@ -1824,7 +1824,18 @@ class Conversation:
         elif message.get("tool_call_id") is not None:
             return False
         tool_calls = message.get("tool_calls")
-        if tool_calls is not None and tool_calls != []:
+        if (
+            (
+                tool_calls is not None
+                and
+                tool_calls != []
+            )
+            and
+            not message.get("content")
+        ):
+            # If tool call request message
+            # doesn't have meaningful content,
+            # we don't serialize it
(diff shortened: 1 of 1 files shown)


src/writer/ai.py
```diff
@@ -1824,7 +1824,18 @@ def _is_serialized(self, message: 'Conversation.Message') -> bool:
         elif message.get("tool_call_id") is not None:
             return False
         tool_calls = message.get("tool_calls")
-        if tool_calls is not None and tool_calls != []:
+        if (
```
FabienArcellier, line 1827: Is there a reason to format like that instead of just doing what follow ?

```python
if tool_calls is not None and tool_calls != [] and not message.get("content"):
    # If tool call request message
    # doesn't have meaningful content,
    # we don't serialize it
    return False
```

mmikita95: I wanted to separate those checks semantically, to indicate that they are the two components for a successful validation - "tool calls are present" and "no content is available". But you're right to point this out, now that I look at it, it looks redundant. Probably this should be even `if not tool_calls and not message.get('content')`.

madeindjs: My two cents: I think it's more readable to introduce a meaningful variable which document the code itself. It might even make the comments not needed.
```py
has_tool_calls = tool_calls is not None and tool_calls != []

if has_tool_calls and not message.get("content"):
    return False
```

mmikita95:
I've decided to go for a simpler approach – as we need to display all messages with proper content, it should be the main "reason" behind the decision to include it into serialization. And, as long as there is no proper content, we shouldn't really care about other properties of the message, it doesn't get displayed either way.

FabienArcellier (approved):

src/writer/ai.py
```diff
@@ -1820,11 +1820,12 @@ def _is_serialized(self, message: 'Conversation.Message') -> bool:
         :return: Boolean that indicates
```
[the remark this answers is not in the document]
mmikita95: Ah, sorry – didn't notice it was cut off. I'll fix it

PR #626, opened 2024-11-11
feat: prevent writer framework version mismatch between local & deploy - WF-67

FabienArcellier:
* when using `writer deploy`, it check the version set in poetry during a cloud deployment is the one used in development

image

* when using `writer run`, it show a warning when the version used is obsolete regards to the development version

image

* when using `writer edit`, it show a warning when the version used is obsolete regards to the development version

image


src/writer/wf_project.py
```diff
@@ -117,13 +127,180 @@ def migrate_obsolete_ui_json(app_path: str, metadata: MetadataDefinition) -> Non
...
+
+def compare_semantic_version(version1: SemanticVersion, version2: SemanticVersion) -> int:
+    """
+    Compares two versions that support semantic version (major, minor, fix, rc).
+
+    * 1 when version1 > version2
+    * 0 when version1 == version2
+    * -1 if version1 < version2
+
+    >>> compare_semantic_version((1, 0, 0, None), (1, 0, 0, 1)) # 1
```
FabienArcellier, line 182: to update

ramedina86:
Hey, I didn't read the whole PR, but relying only on the version pinned in `requirements.txt` or `pyproject.toml` is problematic. And I think that's the case.

As we discussed with @raaymax `requirements.txt` will be optional. Also, there should be no need to pin a Framework version. We should assume that if nothing is specified the latest stable version in PyPi is used.

What I was planning, also why I added

> include the version in the front end for easier troubleshooting (probably not visible, but easily checked via chrome Dev tools/Inspect Element)

is... serve the version via an endpoint, e.g. myapp.com/metadata

then compare that during deployment, AFTER the app has been deployed and issue a warning if it's not the same. I guess the main disadvantage of this it's that it's Writer-specific, but I think it covers this need very well and gives you a much needed way of checking the actual Framework version from the frontend.

ramedina86:
For version comparison the best option seems to be https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

FabienArcellier:
> For version comparison the best option seems to be https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

I will move to it

FabienArcellier:
> then compare that during deployment, AFTER the app has been deployed and issue a warning if it's not the same. I guess the main disadvantage of this it's that it's Writer-specific, but I think it covers this need very well and gives you a much needed way of checking the actual Framework version from the frontend.

Currently it show a warning in the application logs. I will look for a way to show the warning in `writer deploy` if it happens.

ramedina86:
Please make sure to get the effective version and not search for it on `requirements.txt` or `pyproject.toml`, otherwise I'm not too sure how we can cover the scenario where `writer` version isn't specified.

FabienArcellier:
> Please make sure to get the effective version and not search for it on `requirements.txt` or `pyproject.toml`, otherwise I'm not too sure how we can cover the scenario where `writer` version isn't specified.

I search for it only for one specific edge case (poetry on cloud deploy) to give the user the best recommandation. In the other cases, It rely on the comparison package version (`VERSION`) and `./wf/metadata.json`.

ramedina86:
Ah sorry, that's great!

apps/default/poetry.lock
```diff
@@ -0,0 +1,1209 @@
+# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
FabienArcellier, line 1: to remove

src/writer/ss_types.py
```diff
@@ -1,3 +1,4 @@
+import dataclasses
```
FabienArcellier, line 1: rollback those changes

src/writer/deploy.py
```diff
@@ -47,6 +52,16 @@ def deploy(path, api_key, env, verbose, force):
     if not os.path.isdir(abs_path):
         raise click.ClickException("A path to a folder containing a Writer Framework app is required. For example: writer cloud deploy my_app")
 
+    ignore_poetry_version_check = int(os.getenv("IGNORE_POETRY_VERSION_CHECK", '0')) == 1
```
FabienArcellier, line 55: add `WRITER_WF_` , check if the variable is present instead of checking value

ramedina86:
Can this be merged without any changes in the Writer's deployment service repo?

raaymax (approved):

PR #627, opened 2024-11-11
feat: improve performance in writing wf project - WF-97

FabienArcellier:
Since the IDE information is stored in the .wf folder instead of ui.json, typing quickly in the WF IDE may experience latency.

---

The saving of editing files is carried out asynchronously in a dedicated process. It is no longer blocking. The time spent in the process is reduced to practically zero when making modifications in edit mode.

before (extract using py-spy), the operation of write_files use most of the CPU time.
image

after, the operation of write_files is not visible anymore. the CPU time consume by websocket stream is very low
image

### DIFF
--- a/src/writer/wf_project.py
+++ b/src/writer/wf_project.py
@@ -1,16 +1,20 @@
 """
 This module manipulates the folder of a wf project stored into `wf`.
 
->>> wf_project.write_files('app/hello', metadata={"writer_version": "0.1" }, components=...)
+>>> wf_project.write_files_async('app/hello', metadata={"writer_version": "0.1" }, components=...)
 
 >>> metadata, components = wf_project.read_files('app/hello')
 """
+import dataclasses
 import io
 import json
 import logging
+import multiprocessing
 import os
+import time
 import typing
 from collections import OrderedDict
+from multiprocessing import Queue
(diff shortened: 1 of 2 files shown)


src/writer/wf_project.py
```diff
@@ -112,7 +149,7 @@ def migrate_obsolete_ui_json(app_path: str, metadata: MetadataDefinition) -> Non
     file_payload = parsed_file
     # metadata = file_payload.get("metadata", {})
     components = file_payload.get("components", {})
-    write_files(app_path, metadata, components)
+    write_files_async(app_path, metadata, components)
```
FabienArcellier, line 152: rollback this change

src/writer/wf_project.py
```diff
@@ -43,6 +67,31 @@ def write_files(app_path: str, metadata: MetadataDefinition, components: Dict[st
...
 
 
+def start_process_write_files_async(context: WfProjectContext, save_interval: float) -> None:
+    """
+    Creates a process that writes the .wf/ files
+
+    This agent allows you to process the backup of .wf files in the background
+    without blocking application requests.
+
+    :param wf_project_save_interval: the interval in seconds to save the project files
```
FabienArcellier, line 77: rename this parameter

PR #637, opened 2024-11-18
chore: clean up core - WF-80

FabienArcellier:
core refactoring to improve code structure

* move functions that start with `writer_event_handler_*` into an identified class
* move the classes and functions dedicated to the dataframe into the `core_df` module


ramedina86:
Awesome, thank you for this 🙏🏻

PR #641, opened 2024-11-20
fix(ui): display loader during async loading component

madeindjs:
I recently introduced Async Components, but some of them are massive, and the loading time is noticeable. It renders as blank area, which is not good.

So I used `loadingComponent` property to fill the space during the loading, and improved a bit the loading strategy to improve the UX.

https://github.com/user-attachments/assets/1b817790-03d9-4982-8a95-bf1f1ef70ff2

### DIFF
--- a/src/ui/src/builder/BuilderSettings.vue
+++ b/src/ui/src/builder/BuilderSettings.vue
@@ -68,14 +68,19 @@
 
 <script setup lang="ts">
 import { marked } from "marked";
-import { inject, computed, ref, watch } from "vue";
+import { inject, computed, ref, watch, defineAsyncComponent } from "vue";
 import injectionKeys from "../injectionKeys";
 
-import BuilderSettingsHandlers from "./BuilderSettingsHandlers.vue";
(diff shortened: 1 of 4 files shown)


src/ui/src/builder/BuilderSettings.vue
```diff
@@ -68,14 +68,19 @@
...
 import injectionKeys from "../injectionKeys";
 
-import BuilderSettingsHandlers from "./BuilderSettingsHandlers.vue";
 import BuilderSettingsProperties from "./BuilderSettingsProperties.vue";
 import BuilderSettingsBinding from "./BuilderSettingsBinding.vue";
 import BuilderSettingsVisibility from "./BuilderSettingsVisibility.vue";
 import BuilderCopyText from "./BuilderCopyText.vue";
+import BuilderAsyncLoader from "./BuilderAsyncLoader.vue";
+
+const BuilderSettingsHandlers = defineAsyncComponent({
```
madeindjs, line 80: The `BuilderSettingsHandlers` contains the monaco editor, so we don't want to wait everything to be loaded to display the settings.

PR #642, opened 2024-11-21
fix: the process of saving project information stops properly in any condition - WF-114

FabienArcellier:
The process can be stuck or stop with an exception. I migrate it to use `multiprocessing.Event` to trigger stop from inside and avoid killing from outside without control.

image

image


ramedina86:
Yes seems cleaner and more reliable 👍

ramedina86 (approved):

PR #652, opened 2024-11-26
feat(ui): implement `CoreProgressBar` - WF-60

madeindjs:


src/ui/src/composables/useFormater.ts
```diff
@@ -0,0 +1,18 @@
...
+	> = {
+		minimumFractionDigits: 0,
+		maximumFractionDigits: 1,
+	},
+) {
+	const formatter = new Intl.NumberFormat(undefined, {
+		style: "percent",
+		...options,
+	});
+	return computed(() => formatter.format(number.value));
```
ramedina86, line 17: `Formatter` should have two ts, like this `const`

ramedina86:
I don't understand `useFieldValue.ts`, the evaluated fields already come in the expected type. For example, I checked in `CoreSliderInput` the `typeof fields.maxValue.value` and it's number... so not sure why this is needed?

PR #653, opened 2024-11-27
fix: change `object` parameter for `Chat` to comply with Pydantic validation

mmikita95:
After SDK update tests AI module tests began to fail because of this validation. This fix addresses that by providing correct `object` parameter.


ramedina86:
@mmikita95 have we switched the SDK version in the `toml` though?

mmikita95:
Done! Not sure whether we really need this in this particular case, but let's do it nevertheless 😄

ramedina86:
I'd just like to make sure your environment where you're developing matches the `toml`, just to reduce surprises!

PR #655, opened 2024-11-27
chore(ui): improve hash parsing + setup unit test - WF-72

madeindjs:
It's possible to use `URLSearchParams` instead of complex Regex. As it's a very testable part of the code, it was a good opportunity to introduce Vitest and cover it.

This improvement made me realise that a bug was present, and we didn't serialize the params when converting back the object to hash (`.key = length` was not called).

### DIFF
--- a/src/ui/src/core/navigation.ts
+++ b/src/ui/src/core/navigation.ts
@@ -1,46 +1,37 @@
-type ParsedHash = {
+export type ParsedHash = {
 	pageKey?: string;
 	routeVars: Map<string, string>; // Stored as Map to avoid injection e.g. prototype pollution
 };
 
 // Use to communicate with the backend
-type SerializedHash = {
+export type SerializedHash = {
 	pageKey?: string;
 	routeVars: Record<string, string>;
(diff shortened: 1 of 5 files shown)


src/ui/src/core/navigation.ts
```diff
@@ -60,29 +51,23 @@ export function serializeParsedHash(): SerializedHash {
...
-function setHash(parsedHash: ParsedHash) {
+export function serializeParsedHashAsString(parsedHash: ParsedHash) {
 	const { pageKey, routeVars } = parsedHash;
 
 	let hash = "";
 	if (pageKey) {
 		hash += `${encodeURIComponent(pageKey)}`;
 	}
 
-	if (routeVars.keys.length > 0) {
```
madeindjs, line 71: This was a bug, since we need to use `.size`

```js
> const map = new Map()
undefined
> map.keys.length
0
> map.set('a','b')
Map(1) { 'a' => 'b' }
> map.keys.length
0
> map.size
1
```

PR #656, opened 2024-11-28
chore: increase github action runner size

FabienArcellier:
github actions runners sometimes failed due to lack of memory. Activation of superior runner for end 2 ends tests and for tests on macos.


raaymax (approved):
LGTM

PR #660, opened 2024-11-29
fix: patch with fake_generate_content instead of replacing the original

mmikita95:
Replacing the original functions led to fails on consecutive tests. This avoids it, patching the function in the context of the test.


ramedina86:
Thanks for fixing my mess 🙏🏻 will learn from it

PR #666, opened 2024-12-02
feat: `question` endpoint usage for `writer.ai`

mmikita95:
Enables usage of API `question` endpoint via `ask()` / `stream_ask()` methods.
Module-wide methods allow to use multiple instances of `Graph` class or multiple graph IDs to produce an answer; using methods on an instance of `Graph` class allows to query a single graph related to the instance.


ramedina86:
Hey can you explain the `stale` logic and why it's needed?

mmikita95:
Sure! `Graph` objects don't sync automatically with the API. They keep their retrieved state until you manually update them. To handle changes, you'll need logic to mark objects as stale. This ensures that when accessing properties, the module requests updates from the API to retrieve the latest object.

PR #670, opened 2024-12-02
fix(ui): fix builder event selector label

madeindjs:
The `label` and `value` were inverted in the new event selector.


ramedina86:
Hey don't worry I already fixed this, also noticed the `$goToPage` and fixed that too 🙏🏻 thank you though

PR #674, opened 2024-12-03
Fix: Correct image path from 'images/Quickstart' to 'images/quickstart

ajot:


samjulien:
Thanks @ajot! @ramedina86 for the next release, fixing the Quickstart images.

PR #677, opened 2024-12-04
feat: Global events

ramedina86:
This PR is preparation for async jobs. This feature will allow an API endpoint such as `POST /api/job/my_workflow` and later get the result via `GET /api/job/3`.

To facilitate triggering events at this level, this PR introduces "global events" which are `WriterEvent` that don't target a specific component. Therefore, turning `instancePath` into an optional field. Instead, a new `handler` field can be specified to run arbitrary event handlers.

There's of course potential for abuse, so that's why only global events marked as "safe" can be executed. Safe events are produced in "edit" mode. In an upcoming PR, the creation of async jobs will also trigger safe events to run workflows.

### DIFF
--- a/src/writer/serve.py
+++ b/src/writer/serve.py
@@ -48,9 +48,26 @@ MAX_WEBSOCKET_MESSAGE_SIZE = 201*1024*1024
 logging.getLogger().setLevel(logging.INFO)
 
 
+class JobVault:
+
+    def __init__(self):
+        self.counter = 0
+        self.vault = {}
+
+    def generate_job_id(self):
+        self.counter += 1
+        return self.counter
+
+    def set(self, job_id: str, value: Any):
+        self.vault[job_id] = value
+    
+    def get(self, job_id: str):
+        return self.vault.get(job_id)
+
 class WriterState(typing.Protocol):
     app_runner: AppRunner
     writer_app: bool
+    job_vault: JobVault
     is_server_static_mounted: bool
     meta: Union[Dict[str, Any], Callable[[], Dict[str, Any]]] # meta tags for SEO
     opengraph_tags: Union[Dict[str, Any], Callable[[], Dict[str, Any]]] # opengraph tags for social networks integration (facebook, discord)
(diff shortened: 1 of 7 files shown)


tests/backend/test_app_runner.py
```diff
@@ -209,6 +209,37 @@ async def test_bad_event_handler(self, setup_app_runner) -> None:
...
+            assert ev_res.status == "ok"
+            assert not ev_res.payload.result.get("ok")
+
+    @pytest.mark.asyncio
+    @pytest.mark.usefixtures("setup_app_runner")
+    async def test_safe_global_event(self, setup_app_runner) -> None:
+        with setup_app_runner(test_app_dir, "run", load = True) as ar:
+            await init_app_session(ar, session_id=self.proposed_session_id)
+            ev_req = EventRequest(type="event", payload=WriterEvent(
+                type="wf-built-run",
```
ramedina86, line 233: For the moment, `type` in `WriterEvent` isn't being used when specifying a global event (event with no `instancePath`).

src/writer/serve.py
```diff
@@ -122,6 +139,7 @@ async def lifespan(asgi_app: FastAPI):
     """
     app.state.writer_app = True
     app.state.app_runner = app_runner
+    app.state.job_vault = JobVault()
```
ramedina86, line 142: Job Vault will be used to store results of async jobs.

A basic dict-based structure for single-node deployments. There'll be an option to use Redis for bigger deployments (we can then use this same Redis to do other interesting stuff).

FabienArcellier: I have one concern. I don't know how to save the result of the job in the job vault. 

If I understand correctly the goal is to create an API endpoint for triggering a job decoupled from a user session. WF retrieves it and executes it on an `Event handler` of the `AppProcess` 

In this case, the job vault will serve as memory. When the job is finished, it will allow the user to retrieve the result. It is only used for "synchronous" jobs for which we are interested in the result.

My concern is the job vault is in the primary process, the thread pool that run a job in async in WF is in the secondary process. You can write something into the job vault from the secondary process. I have tried to explain what I have understand in the schema below.

image

FabienArcellier: Ok, you rely on ``loop.create_task``. That was the part I wasn't sure. Looks ok for me then.

image

```python
loop = asyncio.get_running_loop()
        task = loop.create_task(app_runner.handle_event(
            session_id, WriterEvent(
                type="wf-builtin-run",
                isSafe=True,
                handler=f"$runWorkflow_{workflow_key}"
            )))
```

PR #683, opened 2024-12-09
fix(ui): add default value for Slider

madeindjs:
The default value for `useFormValueBroker` is an empty string. This make some incompatibilities with components that expect a different type of value and produces props warning.

So I added a params `defaultValue` to `useFormValueBroker` to specifiy the default value when it's needed.


ramedina86:
Great thinking, this is the way 💪🏻

PR #685, opened 2024-12-09
feat: Support icon alias

ramedina86:


src/ui/src/components/workflows/abstract/WorkflowsNode.vue
```diff
@@ -154,6 +142,40 @@ function handleOutMousedown(ev: DragEvent, outId: string) {
...
 }
 
+async function checkIfUrlExists(url: string) {
+	try {
+		const response = await fetch(url, { method: "HEAD" });
+		return response.ok;
+	} catch {
+		return false;
+	}
+}
```
madeindjs, line 152: Don't you want to cache it to avoid sending the same request for every same node ?

It could be done easily by moving this in a separate TS file

```ts
const cacheUrlExists = new Map<string, boolean>();

export async function checkIfUrlExists(url: string) {
    const cache = cacheUrlExists.get(url);
    if (cache !== undefined) return cache;

	try {
		const response = await fetch(url, { method: "HEAD" });
        cacheUrlExists.set(url, true);
		return response.ok;
	} catch {
        cacheUrlExists.set(url, false);
		return false;
	}
}
```

madeindjs: Feel free to assign me a task and I'll take care of it

Issue #578, opened 2024-10-05
Set cookies?


Xynonners:
Hi,

This framework is great (most ergonomic by far of the pure python ones I've tested), but I'm trying to find a way to modify cookies and store data locally on the client side. There doesn't seem to be a properly supported method of doing this currently?

I attempted to mutate the cookies in the session dictionary, but the changes were not propagated to the client side.

Is there any way to do this? Thanks.

FabienArcellier:
We discussed allowing information to be stored in the browser a few months ago. We have ruled out writing cookies.

WF initiates the connection then the exchanges go through a websocket connection. **a cookie cannot be written to the browser**. The table in session is an image of cookies. To write a new cookie, the server need to answer an http request. There is none between the backend and the frontend.

We discussed the possibility of writing to *local storage*. One possibility to be useful would be for the browser to send the content of local storage to the server during initialization and to be available in the `session` attribute.

![image](https://github.com/user-attachments/assets/67061206-f752-4435-89cd-68d16e45ff4b)

Example of API

```python
def on_app_init(state, session):
    state['user_settings']['bgcolor'] = session.local_storage.get('ui_settings', {}).get('bg_color', 'white')

def configure_ui_settings(state, payload, session):
    ui_settings = session.local_storage['ui_settings']
    ui_settings['bg_color'] = payload
    state.set_local_storage('ui_settings', ui_settings) # accept only serializable type


def reset_ui_settings(state, session):
    state.unset_local_storage('ui_settings')

# API
session.local_storage # dict with the content of wf_store from local_storage
session.set_local_storage # set a new value on localstorage
session.unset_local_storage # unset a key from localstorage
```

Xynonners:
> We haven't developped anything about this currently.
> 
> We discussed allowing information to be stored in the browser a few months ago. We have excluded writing cookies due to WF design.
> 
> WF initiates the connection then the exchanges go through a websocket connection. **a cookie cannot be written to the browser**. The table in session is an image of cookies. To write a new cookie, the server need to answer an http request. There is none between the backend and the frontend.
> 
> We discussed the possibility of writing to _local storage_. One possibility to be useful would be for the browser to send the content of local storage to the server during initialization and to be available in the `session` attribute.
> 
[...]

PR #687, opened 2024-12-11
feat/implement local storage persistence - WF-93

FabienArcellier:
The application can save values ​​in local storage and retrieve them when the user reloads their page. This pattern allows you to introduce persistence beyond the current session.

implement #578
---

Peek 2024-12-11 08-37

### DIFF
--- a/src/ui/src/core/index.ts
+++ b/src/ui/src/core/index.ts
@@ -60,12 +60,20 @@ export function generateCore() {
 		addMailSubscription("pageChange", (pageKey: string) => {
 			setActivePageFromKey(pageKey);
 		});
-		addMailSubscription("localStorageSetItem", (event) => {
-			localStorage.setItem("wf." + event.key, JSON.stringify(event.value));
-		});
-		addMailSubscription("localStorageRemoveItem", (event) => {
-			localStorage.removeItem("wf." + event.key);
-		});
+		addMailSubscription(
+			"localStorageSetItem",
+			(event: LocalStorageSetItemEvent) => {
+				localStorage.setItem(event.key, event.value);
+			},
+		);
+
+		addMailSubscription(
+			"localStorageRemoveItem",
+			(event: LocalStorageRemoveItemEvent) => {
+				localStorage.removeItem(event.key);
+			},
+		);
+
 		sendKeepAliveMessage();
 		if (mode.value != "edit") return;
 	}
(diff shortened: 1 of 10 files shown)


src/ui/src/core/index.ts
```diff
@@ -69,6 +75,15 @@ export function generateCore() {
...
 	 */
 	async function initSession() {
+		const localStorageItems = {};
+		for (let i = 0; i < localStorage.length; i++) {
+			const key = localStorage.key(i);
+			if (key.startsWith("wf.")) {
+				const value = localStorage.getItem(key);
+				localStorageItems[key.replace("wf.", "")] = JSON.parse(value);
+			}
+		}
```
madeindjs, line 85: If one value is not valid JSON, it'll throw and stop the execution of the `initSession`. It might be better to catch it and remove the entry if so.

It could be something like

```js
Object.entries(localStorage)
  .filter(([key]) => key.startsWith("wf."))
  .reduce((acc, [key, value]) => {
    try {
      acc[key.replace('wf.', '')] = JSON.parse(value)
    } catch {
      localStorage.removeItem(key)
    }
  }, {});
```

src/ui/src/core/index.ts
```diff
@@ -59,6 +59,12 @@ export function generateCore() {
 		addMailSubscription("pageChange", (pageKey: string) => {
 			setActivePageFromKey(pageKey);
 		});
+		addMailSubscription("localStorageSetItem", (event) => {
+			localStorage.setItem("wf." + event.key, JSON.stringify(event.value));
+		});
+		addMailSubscription("localStorageRemoveItem", (event) => {
+			localStorage.removeItem("wf." + event.key);
+		});
```
madeindjs, line 67: You can add the type
```suggestion
		addMailSubscription("localStorageSetItem", (event: LocalStorageSetItemEvent) => {
			localStorage.setItem("wf." + event.key, JSON.stringify(event.value));
		});
		addMailSubscription("localStorageRemoveItem", (event: LocalStorageRemoveItemEvent) => {
			localStorage.removeItem("wf." + event.key);
		});
```

src/writer/core.py
```diff
@@ -153,6 +157,19 @@ def __init__(self, session_id: str, cookies: Optional[Dict[str, str]], headers:
     def update_last_active_timestamp(self) -> None:
         self.last_active_timestamp = int(time.time())
 
+    def process_local_storage_changes(self) -> None:
```
FabienArcellier, line 160: not required anymore

PR #657, opened 2024-11-28
feat(ui): use design system in builder - WF-127

madeindjs:
- Use `WdsTextField` and `WdsFieldWrapper` in the builder to respect the design system.
- Adapt the design on settings fields to make it compliant

| before | after |
|---|---|
| image | image |
| image  | image |
| image | image |

PR #694, opened 2024-12-16
fix(ui): re-introduce fields unit in Builder - WF-127

madeindjs:
Restore the field unit removed in https://github.com/writer/writer-framework/pull/657

| before | after |
|---|---|
| image | image |

### DIFF
--- a/src/ui/src/builder/settings/BuilderSettingsProperties.vue
+++ b/src/ui/src/builder/settings/BuilderSettingsProperties.vue
@@ -28,7 +28,7 @@
 							: fieldValue.name ?? fieldKey
 					"
 					:hint="fieldValue.desc"
-					:type="fieldValue.type"
+					:unit="fieldValue.type"
 				>
 					<BuilderFieldsColor
 						v-if="fieldValue.type == FieldType.Color"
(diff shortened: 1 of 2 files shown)


ramedina86:
I think we should do same `font-size` for name and type. Inspiration is TypeScript e.g. `ev: WheelEvent`.

madeindjs:
> I think we should do same `font-size` for name and type. Inspiration is TypeScript e.g. `ev: WheelEvent`.

[Implemented](https://github.com/writer/writer-framework/compare/de16adf584b4da130f42c39ea9e73a08681a6061..51b2237970198e657c42ab8640cd9d5cb0600571)

ramedina86:
Is `unit` the proper name of a type I don't know about? 😀

madeindjs:
> Is `unit` the proper name of a type I don't know about? 😀

I found `type` a bit too generic.

Since the `WdsFieldWrapper` is a WDS component, I thought it could be better to have a more meaningful name like `unit`. For example, we could reuse for padding fields.

image

But I won't fight to keep `unit`, I can use `type` if you prefer :grin:

ramedina86:
Ah I see, nah `unit` is fine, not the most descriptive name but I feel the pain with `type` - everything has a `type`.

PR #696, opened 2024-12-16
chore(ui): use WDS colors tokens - WF-149

madeindjs:
Introduce the WDS tokens as CSS variable and use it in the application. The purpose is to uniformize and spread the usage of Writer Design System and prepare the dark mode.

It's mainly a big "search and replace" from HEX colors to CSS vars. But I also replaced some non-WDS colors by the closest equivalent (when it was almost no noticeable).

No big design changes are expected.


ramedina86:
You can also do a forEach and just dump them in a `style`, since these will be injected at the very top and not changed downstream.

madeindjs:
> You can also do a forEach and just dump them in a `style`, since these will be injected at the very top and not changed downstream.

@ramedina86 , good idea! I implemented it in https://github.com/writer/writer-framework/pull/696/commits/3ff5303a0f602d1813c1bdda48768b34b1687d17

PR #700, opened 2024-12-17
feat(ui): Improve wds builder

madeindjs:
Take in account [feedbacks](https://www.figma.com/board/ujJsOLjc2aPSwA7lSclTUC/UI-feedback?node-id=0-1&p=f&t=sSjPKxHnQsj13Aoh-0) after the https://github.com/writer/writer-framework/pull/657

- use `WdsButton` in builder
- avoid rendering empty sections in builder
- set `BuilderSelect` WDS focus state
- set `BuilderSelect`'s dropdown vertical only
- improve `WdsDropdownMenu` search bar

image

### DIFF
--- a/src/ui/src/wds/WdsFieldWrapper.vue
+++ b/src/ui/src/wds/WdsFieldWrapper.vue
@@ -7,15 +7,18 @@
 					>&nbsp;:&nbsp;{{ unit }}</span
 				>
 			</label>
-			<button
+			<WdsButton
 				v-if="helpButton"
 				class="WdsFieldWrapper__title__help"
-				variant="subtle"
-				:title="typeof helpButton === 'string' ? helpButton : undefined"
+				variant="neutral"
+				size="smallIcon"
+				:data-writer-tooltip="
+					typeof helpButton === 'string' ? helpButton : undefined
+				"
 				@click="$emit('helpClick')"
 			>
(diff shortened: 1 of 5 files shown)


src/ui/src/wds/WdsFieldWrapper.vue
```diff
@@ -7,22 +7,27 @@
...
 				>
 			</label>
-			<button
+			<WdsButton
 				v-if="helpButton"
 				class="WdsFieldWrapper__title__help"
-				variant="subtle"
-				:title="typeof helpButton === 'string' ? helpButton : undefined"
+				variant="neutral"
+				size="smallIcon"
```
madeindjs, line 14: @ramedina86 , about the help button, the issue is: The help icon includes the circle ([here](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:help:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=help&icon.size=24&icon.color=%23e8eaed)), thus `tertiary` variant feels weird

image

I thought about using `question_mark` icon instead, but it's still not that pretty

image


So I decided to stik to `neutral` and this is the result.

https://github.com/user-attachments/assets/4d42933e-c4f8-4fd7-9d59-0735f96e3650

WDYT ?

PR #702, opened 2024-12-17
feat(ui): use `WdsButton` for `BaseCollapseButton` - WF-132

madeindjs:
The `BaseCollapseButton` used a custom button with a different `:focus` state. I moved it to the standard `WdsButton` and implemented a `:focus-visible` to it to improve tab navigation.

The change should be almost transparent in term of UI.

https://github.com/user-attachments/assets/80159bd3-346a-4a16-b2f4-e91675c26c89


ramedina86:
Let's change the outline for now

madeindjs:
> Let's change the outline for now

It's done 👍

PR #704, opened 2024-12-18
chore(ui): introduce a logger proxy - WF-85

madeindjs:
Introduce a simple abstraction to use logger in the application. For the moment, it's just a proxy to `console`, but it can be plugged to any library later. It's also a good way to prevent that logs are not displayed in the tests.


src/ui/src/composables/useLogger.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-console */
+
+/**
+ * A simple abstraction to use logger in the application. For the moment, it's just a proxy to `console`, but it can be plugged to any library later.
+ */
+export function useLogger(): Pick<
```
FabienArcellier, line 6: What the interest of using a hook syntax here and not use a function as ?

```
logger.log("my logger")
```

instead of 

```
useLogger().log()
```

PR #706, opened 2024-12-18
fix(ui): handle long text in `BuilderSelect`

madeindjs:
The `BuilderSelect` trigger has overflow issues with long text containing breakable content. The trick was just a missing `white-space: nowrap` rule.

I also took the opportunity to add a tooltip to the `BuilderSelect`'s trigger.

Screenshot 2024-12-18 at 19 15 29

### DIFF
--- a/src/ui/src/builder/BuilderSelect.vue
+++ b/src/ui/src/builder/BuilderSelect.vue
@@ -3,13 +3,18 @@
 		<button
 			class="BuilderSelect__trigger"
 			role="button"
-			:data-writer-tooltip="currentLabel"
 			@click="isOpen = !isOpen"
 		>
 			<i v-if="!hideIcons" class="material-symbols-outlined">{{
 				currentIcon
 			}}</i>
-			<div class="BuilderSelect__trigger__label">{{ currentLabel }}</div>
+			<div
+				class="BuilderSelect__trigger__label"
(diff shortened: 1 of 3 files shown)


FabienArcellier:
I add this issue 2 weeks ago. It would be awesome to have the popup on the list after 3 seconds to know what is the handler. do you think it would be possible ?

image

madeindjs:
Good point @FabienArcellier . I thought about it, it's easy as adding an attribute `data-writer-tooltip` on each option. The issue is, the tooltip will be displayed every time.

I would like to display the tooltip only if the element overflow. Maybe adding a new prop `data-writer-tooltip-strategy="overflow"`.

I can give a try, it should not be that hard to implement this logic here

https://github.com/writer/writer-framework/blob/4e956b15bf49c511c44936b4ddc359ea2d780759/src/ui/src/builder/BuilderTooltip.vue#L72-L76

madeindjs:
@FabienArcellier , I implemented it https://github.com/writer/writer-framework/pull/706/commits/2a2f2d2e461188e721c6f7af93448402ecd8f392 . So we display the tooltip in `BuilderSelect` only if the element overflow

https://github.com/user-attachments/assets/5dc25e10-7cfc-4612-b4c5-044522290818

PR #709, opened 2024-12-23
chore(ui): improve icon fetching. WF-141

madeindjs:
Introduce a new component `ShareImgWithFallback` which takes an array of URLs for the image and uses the first successful URL.

This component uses a new hook `useAssetContentType` that takes care of fetching the `Content-Type` header of the asset URL, and handles the concurrency / cache. The `Content-Type` is needed because the server might serve a default `text/html` if the assets doesn't exist.

It also fix an edge case wherein the application run is an URL path like `https://streamsync-dev.osc-fr1.scalingo.io/hello/`, we are trying to import asset from the root (`https://streamsync-dev.osc-fr1.scalingo.io/`) and not the path `/hello`

I also took the opportunity to test the Vue.js component using [Vue Test Utils](https://test-utils.vuejs.org/)

This change is transparent for the user.

### DIFF
--- a/src/ui/src/builder/sidebar/BuilderSidebarToolkit.vue
+++ b/src/ui/src/builder/sidebar/BuilderSidebarToolkit.vue
@@ -23,19 +23,10 @@
 					@dragend="handleDragEnd($event)"
 					@dragstart="handleDragStart($event, tool.type)"
 				>
-					<img
+					<SharedImgWithFallback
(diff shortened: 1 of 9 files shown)


ramedina86:
Looks good, glad to see the paths fixed too. Can you look into the merge conflict? it's a straightforward one

PR #717, opened 2025-01-03
feat: implement workflow blocks to interact with third tools / airtable

FabienArcellier:
* implement workflow blocks
* implement functional tests


FabienArcellier:
drop for #772

PR #718, opened 2025-01-03
fix(ui): avoid bundling `injectionKeys` in custom components

madeindjs:
The custom component built using `npm run custom.build` was not working correctly because the `provide/inject` failed.

The error was, we bundled `injectionKeys` in the `template.umd.js` , So it makes defining two different `Symbol` which point to a different `provide`.

It's because the `rollupOptions.external` in Vite config pointed to `../injectionKeys`, but we sometime use the alias `@/injectionKeys`. So I simply use `path.resolve` to be sure `injectionKeys` are not bundled in custom components.

### DIFF
--- a/src/ui/vite.config.custom.ts
+++ b/src/ui/vite.config.custom.ts
@@ -4,8 +4,6 @@ import { defineConfig, UserConfig } from "vite";
 import vue from "@vitejs/plugin-vue";
 import writerPlugin from "./viteWriterPlugin";
 
-const injectionKeys = path.resolve("src/injectionKeys");
-
 export default defineConfig({
 	base: "./",
 	plugins: [vue(), writerPlugin()],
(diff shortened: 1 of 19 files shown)


src/ui/vite.config.custom.ts
```diff
@@ -29,11 +31,11 @@ export default defineConfig({
 			},
 		},
 		rollupOptions: {
-			external: ["vue", "../injectionKeys"],
+			external: ["vue", injectionKeys],
```
FabienArcellier, line 34: I have to use this code to make the build of custom component working on windows.

```javascript
rollupOptions: {
			external: ["vue", "@/injectionKeys"],
			output: {
				globals: {
					vue: "vue",
					[path.resolve("src/injectionKeys")]: "injectionKeys",
				},
			},
		},
		```

madeindjs: I implemented it in https://github.com/writer/writer-framework/pull/718/commits/d6fb5ec06ce879305bc244e0fc5919691502968b

PR #720, opened 2025-01-03
feat: support python 3.13

FabienArcellier:
the current version of WF does not support python 3.13 because of a dependency limit on apache arrow (for all platforms) and a dependency constraint on numpy (for windows)

* update dependencies to support of python 3.13 
* update continuous integration pipeline to check WF on python 3.13


madeindjs (approved):
I tested it with Python 3.13.1 and it works:

```sh
cd /tmp
git clone git@github.com:FabienArcellier/writer-framework.git
git checkout enable-python-3-13
poetry install
poetry run alfred install.dev
poetry run writer edit apps/hello --port 5000
```

PR #724, opened 2025-01-08
feat: implement multi-file code editor - WF-100

madeindjs:
Implement the feature to allow the user to create, update and delete any source file of the application.

The logic flow is:

1. Python backend returns the `sourceFiles` in the Init request. It's an object representing the tree structure like
    ```json
    {"type": "directory", "children": {"README.md": {"type": "file", "content": "This app w", "complete": false}}}
    ```
3. the frontend stores this and creates a  copy of this as `sourceFilesDraft`
4. different action in the editor will trigger new sockets events I implemented: `createSourceFile`, `deleteSourceFile` or `renameSourceFile`

The main challenge was the synchronization between `sourceFiles` and `sourceFilesDraft`. This main logic stands in the composable `useSourceFilesDraft`.

## Quick demo

https://github.com/user-attachments/assets/be75f433-5101-4895-a5a6-c51029cafab1

### DIFF
--- a/src/writer/app_runner.py
+++ b/src/writer/app_runner.py
@@ -7,6 +7,7 @@ import multiprocessing
 import multiprocessing.connection
 import multiprocessing.synchronize
 import os
+import shutil
 import signal
 import subprocess
 import sys
@@ -749,7 +750,8 @@ class AppRunner:
         to_path_abs = os.path.join(self.app_path, to_path)
         self._check_file_in_app_path(to_path_abs)
 
-        os.renames(from_path_abs, to_path_abs)
+        os.makedirs(os.path.dirname(to_path_abs), exist_ok=True)
+        os.rename(from_path_abs, to_path_abs)
 
         self.source_files = wf_project.build_source_files(self.app_path)
 
@@ -760,7 +762,10 @@ class AppRunner:
         path = os.path.join(self.app_path, file)
         self._check_file_in_app_path(path)
 
-        os.remove(path)
+        if os.path.isfile(path):
+            os.remove(path)
+        else:
+            shutil.rmtree(path)
 
         self.source_files = wf_project.build_source_files(self.app_path)
 
--- a/src/writer/app_runner.py
+++ b/src/writer/app_runner.py
@@ -8,6 +8,7 @@ import multiprocessing.connection
 import multiprocessing.synchronize
 import os
 import signal
+import subprocess
 import sys
 import threading
 from types import ModuleType
@@ -43,6 +44,7 @@ from writer.ss_types import (
     InitSessionRequestPayload,
     InitSessionResponsePayload,
     ServeMode,
+    SourceFilesDirectory,
     StateContentRequest,
     StateContentResponsePayload,
     StateEnquiryRequest,
@@ -501,9 +503,9 @@ class FileEventHandler(watchdog.events.PatternMatchingEventHandler):
     Watches for changes in files and triggers code reloads.
     """
 
-    def __init__(self, update_callback: Callable):
+    def __init__(self, update_callback: Callable, patterns: List[str]):
         self.update_callback = update_callback
-        super().__init__(patterns=["*.py"], ignore_patterns=[
+        super().__init__(patterns=patterns, ignore_patterns=[
             ".*"], ignore_directories=False, case_sensitive=False)
 
     def on_any_event(self, event) -> None:
@@ -605,6 +607,7 @@ class AppRunner:
         self.client_conn: Optional[multiprocessing.connection.Connection] = None
         self.app_process: Optional[AppProcess] = None
         self.run_code: Optional[str] = None
+        self.source_files: SourceFilesDirectory = {"children": {}, "type": "directory"}
         self.bmc_components: Optional[Dict] = None
         self.is_app_process_server_ready = multiprocessing.Event()
         self.is_app_process_server_failed = multiprocessing.Event()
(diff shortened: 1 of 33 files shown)


FabienArcellier:
I would advocate having a context menu to delete an item. With the current pattern, I can't remove a folder when I inadvertently create one : https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event

image

FabienArcellier:
I would advocate to bind CTRL+S when the user is inside the editor. It currently open the browser windows : https://codepen.io/ThatSoftwareDude/pen/poVQowE

FabienArcellier:
We lost ability to save a file when an error is raised changing the python code. 

image

src/writer/app_runner.py
```diff
@@ -104,6 +106,7 @@ def __init__(self,
         self.app_path = app_path
         self.mode = mode
         self.run_code = run_code
+        self.source_files: SourceFilesDirectory = {"children": {}, "type": "directory"}
```
FabienArcellier, line 109: I think the attribute is not used. You should remote it.

FabienArcellier:
I rename and file in `.wf` directory. It has wrong behavior.

image

src/writer/app_runner.py
```diff
@@ -667,6 +672,53 @@ def load(self) -> None:
...
+                        content = self.load_persisted_script(relative_path)
+                        # limit only the first 100 characters to limit bandwidth usage, the rest will be lazy loaded
+                        exerpt = content[0:100]
+                        current_level["children"][part] = {
+                            "type": "file",
+                            "content": exerpt,
+                            "complete": exerpt == content,
+                        }
+                    except UnicodeDecodeError:
+                        print(relative_path)
```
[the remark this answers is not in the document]
madeindjs: done in https://github.com/writer/writer-framework/pull/724/commits/8de3fe31424b501b13aee6a882ab8a912bee4116

src/writer/app_runner.py
madeindjs: done in https://github.com/writer/writer-framework/pull/724/commits/8de3fe31424b501b13aee6a882ab8a912bee4116

src/writer/app_runner.py
```diff
@@ -667,6 +672,53 @@ def load(self) -> None:
         # parent pid and pid.
         self._subscribe_terminal_signal()
 
+    def _build_source_files(self) -> SourceFilesDirectory:
```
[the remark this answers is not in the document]
madeindjs: done in https://github.com/writer/writer-framework/pull/724/commits/8de3fe31424b501b13aee6a882ab8a912bee4116

madeindjs:
> I would advocate having a context menu to delete an item. With the current pattern, I can't remove a folder when I inadvertently create one : https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event

@FabienArcellier , what do you think about simply avoiding rendering empty folder?

So, if you create a file `a/b/c.txt`, it creates the folder `a/b/`, and when you remove `c.txt`, we don't display the folder in the UI.

The edge case is, when the user tries to create a file matching an empty folder path. In this situation, we can simply delete the empty folder before creating the file.

src/writer/app_runner.py
```diff
@@ -644,14 +648,37 @@ def _set_logger(self):
 
     def _start_fs_observer(self):
         self.observer = PollingObserver(AppRunner.UPDATE_CHECK_INTERVAL_SECONDS)
-        self.observer.schedule(FileEventHandler(self.reload_code_from_saved), path=self.app_path, recursive=True)
+        self.observer.schedule(FileEventHandler(self.reload_code_from_saved, patterns=["*.py"]), path=self.app_path, recursive=True)
+        self.observer.schedule(FileEventHandler(self._install_requirements, patterns=["requirements.txt"]), path=self.app_path)
```
FabienArcellier, line 652: You don't have access to the session here. If you want to install, you should do that in the request. The challenge it's it will take some time.

src/writer/wf_project.py
```diff
@@ -321,3 +322,61 @@ def can_create_project(path: str) -> bool:
...
+    """
+    def load_persisted_script(file: str) -> str:
+        path = os.path.join(app_path, file)
+        with open(path, "r", encoding='utf-8') as f:
+            return f.read()
+
+    # we can't use `glob`'s `include_hidden` options since it's only available on Python 3.11+
+    files = []
+    for pattern in ['**/*', '**/.*', '.**/*', '.**/.*']:
+        files += glob.glob(os.path.join(app_path, pattern), recursive=True)
```
FabienArcellier, line 346: replace with more mechanical pattern

```python
    for root, _, filenames in os.walk(app_path):
        for filename in filenames:
            # Create the relative file path to the app_path
            file_path = os.path.realpath(os.path.join(root, filename))
            files.append(file_path)
```

src/writer/serve.py
```diff
@@ -546,10 +547,36 @@ async def _handle_incoming_edit_message(websocket: WebSocket, session_id: str, r
...
-            app_runner.update_code(
-                session_id, req_message.payload["code"])
+            app_runner.update_code(session_id, req_message.payload["code"])
+        elif req_message.type == "loadSourceFile":
+            path = os.path.join(*req_message.payload['path'])
+            try:
+                response.payload = { "content": app_runner.load_persisted_script(path) }
+            except FileNotFoundError as error:
+                logging.warning(f"could not load script at {path}", error)
+                response.payload = {"error": str(error)}
```
FabienArcellier, line 559: maybe we have to exten the `mail protocol` to get it on all the websocket message

PR #729, opened 2025-01-21
feat(ui): redesign Dataframe editor - AI-23

madeindjs:
Restyle the Dataframe editor component (still behind the flag `dataframeEditor`). The main changes are:

- Hovering a row now displays all editable fields 
- Editable fields use WDS components for input / textarea
- Introduces a new WDS component `WdsNumberInput` ([Figma](https://www.figma.com/design/jgLDtwVwg3hReC1t4Vw20D/.WDS-Writer-Design-System?node-id=1102-2359&t=ixoGoEU4atTvsEEO-0))
- The `display: grid` used for the table now use `subgrid` property, which is more handy
- The "Add row" is now outside the table, and scroll to the added row when clicked
- The resize hand is more visible
- The "sorted" icon is displayed when the column is sorted 

https://github.com/user-attachments/assets/5774b80b-5506-4026-98e4-e9f3653f5294


ramedina86:
Looks awesome. Few conflicts here. Also I think time to remove flag. What do you think?

PR #730, opened 2025-01-24
perf(ui): use Async component to avoid loading ressources - WF-107

madeindjs:
WIP


madeindjs:
I close this POC for now as it requires a big refactoring to export component's fields in another TS file. It's working, but add more complexity. I don't think it worth it since big libraries are already using dynamic import for bundle splinting

PR #731, opened 2025-01-24
feat(ui): support markdown in `AnnotatedText` - WF-169

madeindjs:
The purpose is to support markdown in `CoreAnnotatedText`.

So I introduced a new setting `useMarkown` for this component and:

1. extracted markdown style from `BaseMarkdown` to `BaseMarkdownRaw`
2. constructed the markdown in a function `useMarkdownRenderer` which will construct the content with `<span>` for annotations

Screenshot 2025-01-27 at 17 45 39


ramedina86:
Pretty awesome, this unlocks some big ticket use cases.

Can we have titles? Why isn't "# This" rendering?

ramedina86:
Amazing stuff! :party-parrot:

PR #739, opened 2025-02-11
 feat: setup knwoledge graph selector - WF-138 

madeindjs:
This feature introduces a new selector that allows the user to select a knowledge graph from Writer's platform instead of copy/pasting the UUID.

| in workflow block | in tools editor |
|---|---|
| Screenshot 2025-02-11 at 17 53 43 | Screenshot 2025-02-11 at 17 54 06 |

- introduce a new event called `listResources` that gets the list of Writer's graph (and will support other things later)
- make `BuilderSelect` and `WdsDropdownMenu` compatible with multi-selector
- introduce a new field `FieldType.WriterGraphId` to support graph selector for component's settings


ramedina86:
Nice one, I saw that you're also introducing the local storage management with a composable and the mode remembering.

You have some conflicts, please solve and I'll merge 🙏🏻

PR #746, opened 2025-02-14
feat(ui): improve component tree search. WF-183

madeindjs:
When searching for a component in the tree, we start to display only components matching the query, or those containing children matching the query.

It'll

- collapse component that doesn't match the query (with a small animation)
- display the number of matches
- allow to clear search with a button

Before

https://github.com/user-attachments/assets/0f0ca13e-711d-4861-9cbc-058acdd00012

After

https://github.com/user-attachments/assets/a31a137a-61e6-4811-a513-e536e88e9c40


ramedina86:
Looks so neat!

PR #753, opened 2025-02-23
feat: Announcement queue with mail support

ramedina86:
This PR switches [asyncio.Condition](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition) for [asyncio.Queue](https://docs.python.org/3/library/asyncio-queue.html) when dealing with announcements, providing a more flexible architecture that also allows for announcement payloads to be sent.

This enabled using the announcement payload to send mail, making it possible to send mail when outside the app process. For example, for system notification purposes.

This feature was applied to send `pip install` logs.

### DIFF
--- a/src/writer/app_runner.py
+++ b/src/writer/app_runner.py
@@ -695,7 +695,6 @@ class AppRunner:
                     }
                 ],
             )
-            # TODO(WF-170): find a way to dispatch log
         except Exception as e:
             logger.warning(f"Unexpected error: {e}")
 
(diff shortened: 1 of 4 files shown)


src/writer/app_runner.py
```diff
@@ -672,7 +682,20 @@ def _install_requirements(self) -> None:
             self.reload_code_from_saved()
```
madeindjs, line 682: Do you think we should add a log when the installation is successful?

madeindjs (approved):

src/writer/app_runner.py
ramedina86: Yes, I'll prepare another PR and I'll include this, good idea.

PR #764, opened 2025-02-28
feat(ui): implement link from UI to workflow - WF-203

madeindjs:
Implement a new design for linking UI elements to workflow:

- ⚠️ Remove custom handler and handler using Python function
- Implement a new component `BuilderListItem` to draw a list of elements with an SVG lime
- Add new `action` property to Toast. I used it to display a "go to created workflow" button in the toast
- Add logic to display images in `WdsDropdownMenu`
- Introduce two new field types `Component Id` &  `Component Event Type`

https://github.com/user-attachments/assets/de21da9d-60b1-4789-9c6d-9efd9dece007


ramedina86:
- Let's double check the `isSafe` logic to make sure runBranch (and the original runWorkflow) handlers cannot be run in run mode
- Let's verify there's no displacement when the loader is shown in the blueprint tree

PR #772, opened 2025-03-12
feat: implement group block in agent editor to support vendor block - WF-181

FabienArcellier:
Agent editor blocks can be grouped. The developer customizes the action in the block's context menu.

* Attributes are copied when the block is copied
* The block can be customized at any time

image


src/ui/src/builder/settings/BuilderSettingsProperties.vue
```diff
@@ -240,6 +239,14 @@ function handleExpand(fieldKey: string) {
...
 	expandedFields.value.delete(fieldKey);
 }
+
+interface Props {
+	groupLevel: 0 | 1;
+}
+
+const props = withDefaults(defineProps<Props>(), {
+	groupLevel: 0,
+});
```
madeindjs, line 249: It's better to use object notation to define props
```suggestion
const props = defineProps({
	groupLevel: { type: Number as PropType<0 | 1>, required: false, default: 0 },
});
```

poetry.lock
```diff
@@ -1,11 +1,13 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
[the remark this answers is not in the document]
FabienArcellier: I rollback this part. I first add a new dependency on python-dotenv. As I update my poetry version, lock file has been modified. Finally, I remove the backend_external testing, it's too difficult to maintain and rollback poetry lock and pyproject.toml changes.

src/ui/src/builder/settings/BuilderSettingsActionSelect.vue
```diff
@@ -0,0 +1,118 @@
...
+}
+
+.suggestion-item {
+	padding: 8px;
+	cursor: pointer;
+}
+
+.suggestion-item:hover {
+	background-color: #f0f0f0;
+}
```
madeindjs, line 117: You can remove this dead code

```suggestion
.selector-container {
	position: relative;
}
```

madeindjs (approved):

PR #796, opened 2025-03-26
feat: Improved interpolation

ramedina86:
Improved string interpolation, adjusting to whether the string is being interpreted as json or not
Introduces _HTTP Request_ body type and _Set State_ value type

### DIFF
--- a/src/ui/src/renderer/useFormValueBroker.ts
+++ b/src/ui/src/renderer/useFormValueBroker.ts
@@ -27,6 +27,12 @@ export function useFormValueBroker<T = any>(
 	const component = computed(() => wf.getComponentById(componentId));
 	const { evaluateExpression } = useEvaluator(wf);
 
+	function initializeFormValueBroker() {
+		const bindingEventType = component.value.binding?.eventType;
+		if (!bindingEventType) return;
+		handleInput(defaultValue, bindingEventType);
+	}
+
 	function getBindingValue() {
 		const component = wf.getComponentById(componentId);
 		if (component?.binding?.stateRef) {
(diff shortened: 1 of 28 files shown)


src/ui/src/renderer/useFormValueBroker.ts
```diff
@@ -27,6 +27,12 @@ export function useFormValueBroker<T = any>(
 	const component = computed(() => wf.getComponentById(componentId));
 	const { evaluateExpression } = useEvaluator(wf);
 
+	function initializeFormValueBroker() {
+		const bindingEventType = component.value.binding?.eventType;
+		if (!bindingEventType) return;
+		handleInput(defaultValue, bindingEventType);
+	}
```
madeindjs, line 34: Can we add the `onMounted` logic here to avoid doing it in every component?

```ts
onMounted(() => {
	initializeFormValueBroker();
});
```
I don't get if we always want this behavior. Maybe it can be a parameter of the composition function

```ts
export function useFormValueBroker<T = any>(
	/* ... */
   opts: { initializeDefaultValue?: boolean } = {}
) {
  if (opts?.initializeDefaultValue) {
    onMounted(initializeFormValueBroker);
  }
  // ...
}
```

ramedina86: Makes sense, didn't think about doing the onMounted on the composable but it's definitely cleaner.

I'd say we always want it because it doesn't make much sense to have a text input that's null, then "a" if you type a, then "" if you backspace. And then you have to handle a null in the blueprint which might be problematic if you're sending it in an HTTP request etc.

ramedina86: Changed, thanks for the suggestion, big improvement

PR #797, opened 2025-03-26
fix(ui): handle unknown component in Blueprint - WF-193

madeindjs:
Customize `RenderError.vue` to display a `WorkflowsNode` component for invalid workflow blocks.

It also renders the `outs` of the nodes. 

image

### DIFF
--- a/src/ui/src/components/workflows/WorkflowsWorkflow.vue
+++ b/src/ui/src/components/workflows/WorkflowsWorkflow.vue
@@ -906,30 +906,4 @@ svg {
 	width: 100%;
 	height: 100%;
 }
-
-/* customize the `RenderError` component to make it looks like a node */
-.nodeContainer :deep(.RenderError) {
-	position: absolute;
-	width: 240px;
-	border-radius: 8px;
-	box-shadow: var(--wdsShadowBox);
-	padding: 12px;
-	background-color: var(--builderBackgroundColor);
-	border: 2px solid transparent;
-	font-size: 14px;
-}
-.nodeContainer :deep(.RenderError.selected) {
-	background-color: var(--builderBackgroundColor);
-	border: 2px solid var(--wdsColorBlue4);
-	font-size: 14px;
-}
-.nodeContainer :deep(.RenderError:hover) {
-	border: 2px solid var(--wdsColorBlue2);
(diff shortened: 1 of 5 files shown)


src/ui/src/components/workflows/WorkflowsWorkflow.vue
```diff
@@ -906,4 +906,26 @@ svg {
...
+	border: 2px solid transparent;
+	font-size: 14px;
+}
+.nodeContainer :deep(.RenderError.selected) {
+	background-color: var(--builderBackgroundColor);
+	border: 2px solid var(--wdsColorBlue4);
+	font-size: 14px;
+}
+.nodeContainer :deep(.RenderError h2) {
+	color: var(--wdsColorOrange5);
```
ramedina86, line 927:
```suggestion
	color: var(--builderErrorColor);
```

ramedina86: Same thing though

src/ui/src/components/workflows/WorkflowsWorkflow.vue
```diff
@@ -906,4 +906,26 @@ svg {
...
+	font-size: 14px;
+	font-weight: 600;
```
ramedina86, line 929: Do we use 600 anywhere? is this one of the known Headers?

ramedina86:
Did you look into doing something aligned with `fallbackTemplate` in `templateMap.ts`?

madeindjs:
> Did you look into doing something aligned with `fallbackTemplate` in `templateMap.ts`?

@ramedina86  I'm not sure that it is needed. We display the error component from here

https://github.com/writer/writer-framework/blob/6e9bf12375741cdbf64d9d2160241548e7ca1ce7/src/ui/src/renderer/ComponentProxy.vue#L350-L353

https://github.com/writer/writer-framework/blob/6e9bf12375741cdbf64d9d2160241548e7ca1ce7/src/ui/src/renderer/ComponentProxy.vue#L305

So I'm just customizing the component `RenderError` inside `WorkflowsWorkflow`. It just allows the user to see the broken node and remove it if needed.

Do you think I'm missing something?

PR #802, opened 2025-04-01
Autogen loader

polymorpheuz:
## Description

Basically, added the autogen loader :D

https://github.com/user-attachments/assets/6ca47e52-57db-4c31-b214-3311d4c31f48

### DIFF
--- a/src/ui/src/components/workflows/WorkflowsGenerationLoader.vue
+++ b/src/ui/src/components/workflows/WorkflowsGenerationLoader.vue
@@ -1,0 +1,133 @@
+<template>
+	<div class="WorkflowsLifeLoading">
+		<div class="logoContainer">
+			<img src="../../assets/logo.svg" alt="Writer Framework logo" />
+			<WdsLoaderDots class="dot-loader" color="black" />
+			<div class="rotating-text">
+				<span v-for="(text, index) in texts" :key="index" :class="{ active: currentTextIndex === index }">
+					{{ text }}
+				</span>
+			</div>
+		</div>
+		<div class="background">
+			<RotatingCircle custom-class="rotating-circle left" />
+			<RotatingCircle custom-class="rotating-circle center" />
+			<RotatingCircle custom-class="rotating-circle right" />
+		</div>
+	</div>
+</template>
+
+<script setup lang="ts">
+import { onMounted, ref } from 'vue';
+import WdsLoaderDots from "@/wds/WdsLoaderDots.vue";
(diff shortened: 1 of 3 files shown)


src/ui/src/components/workflows/WorkflowsGenerationLoader.vue
```diff
@@ -0,0 +1,133 @@
...
+	animation: rotateCounterClockwise 12s linear infinite;
+	top: -50%;
+	left: -30%;
+	z-index: 1;
+}
+
+.rotating-circle.center {
+	animation: rotateClockwise 16s linear infinite;
+	top: -50%;
+	left: 17%;
```
polymorpheuz, line 123: Hear me out. For some reason, that I couldn't find yet, this little dude ignored all the translates and I couldn't center it that way. No clue why for now.

polymorpheuz:
@ramedina86 Oh, sorry. Fixed this guy now!

src/ui/src/components/workflows/WorkflowsGenerationLoader.vue
```diff
@@ -0,0 +1,143 @@
...
+			<WdsLoaderDots class="dot-loader" color="black" />
+			<div class="rotating-text">
+				<span
+					v-for="(text, index) in texts"
+					:key="index"
+					:class="{ active: currentTextIndex === index }"
+				>
+					{{ text }}
+				</span>
+			</div>
```
madeindjs, line 14: It's better to use `<p>` for text. We have a reset style, so using `<p>` instead of `<div>` should not affect the style
```suggestion
			<p class="rotating-text">
				<span
					v-for="(text, index) in texts"
					:key="index"
					:class="{ active: currentTextIndex === index }"
				>
					{{ text }}
				</span>
			</p>
```

madeindjs (commented):
Some minor comments but good job 💪

src/ui/src/components/workflows/WorkflowsGenerationLoader.vue
madeindjs: I guess it's because you use `custom-class` to apply the `class` in the children, but `.left` is defined in the parent , and the style is scoped.

Using `class` as I suggested, and moving all the styles in the parent instead might solve your issue.

https://github.com/writer/writer-framework/blob/c6f4544d7284491c7c4d275cbe34badb9e440b88/src/ui/src/components/workflows/WorkflowsGenerationLoader.vue#L17

PR #819, opened 2025-04-10
chore: session context for block execution

mmikita95:
Enables session context for injection of `X-Agent-Token` header during API calls. Modifies the AI module to issue a new client with injected header on each attempt of client retrieval.


ramedina86:
I'm pretty sure you can just use the existing `use_request_context` as used by `_handle_message` (`app_runner.py`).

```
        with use_request_context(session_id, request):
            session = writer.session_manager.get_session(session_id)
            if not session:
                raise MessageHandlingException("Session not found.")
```

Avoid making anything blueprints/block specific. Please also consider autogen.

PR #827, opened 2025-04-15
fix(ui): improve writer agent deployment

madeindjs:
- compute the agent URL on frontend side
- rely on `status` of the agent to detect if the application is deployed or not
- improve types
- fix runtime warning

### DIFF
--- a/src/ui/src/composables/useApplicationCloud.ts
+++ b/src/ui/src/composables/useApplicationCloud.ts
@@ -1,6 +1,14 @@
 import { WriterApi, WriterApiApplicationDeployment } from "@/writerApi";
 import { Core } from "@/writerTypes";
-import { computed, onMounted, readonly, ref, shallowRef, watch } from "vue";
+import {
+	computed,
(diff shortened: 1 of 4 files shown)


src/ui/src/composables/useApplicationCloud.ts
```diff
@@ -31,16 +39,11 @@ export function useApplicationCloud(wf: Core) {
...
-		return new URL(
-			`/aistudio/organization/${orgId.value}/agent/${appId.value}/deploy`,
-			apiBaseUrl,
-		);
-	});
-
 	const liveUrl = computed(() => {
-		return deploymentInformation.value.applicationVersionData.data
-			?.deploymentUrl;
+		const teamId = deploymentInformation.value.writer?.teamIds?.[0];
```
madeindjs, line 43: Need to fall back to  `cloud` team if app is deployed on cloud (`deploymentInformation.value.cloud?.teamIds?.[0]`). Discussion is in progress about that

PR #832, opened 2025-04-17
feat(ui): introduce the App name renamer field - WF-355

madeindjs:
Add an input field in the navbar to update the Agent name through the endpoint

```http
PUT /api/template/organization/{orgId}/application/{appId}/metadata
```

The logic in `useApplicationCloud` will handle synchronization with the special block `root` and update the field `appName`.

image


madeindjs:
superset by https://github.com/writer/writer-framework/pull/889

PR #834, opened 2025-04-22
feat: use "Sentence case" for components names - WF-356

madeindjs:
use "Sentence case" instead of "Title Case" for all components names


madeindjs:
Closed for now since it's a low-priority issue, and it's a breaking change for Low Code UI components.

PR #840, opened 2025-04-28
feat(ui): setup FullStory - WF-368

madeindjs:


madeindjs:
Moving as draft because I'm waiting for Fullstory is activated on writer.com

PR #851, opened 2025-05-01
feat: save file MD5 hash

UladzislauK-Writer:


mmikita95 (approved):

UladzislauK-Writer:
@ramedina86 some context for you from our conversation with Mikita.

First commit implements storing MD5 hash as the first line in the file itself.
Second, as an optimization, stores hashes in memory, in the context dataclass. It was safer to keep the first approach since there is [migrate_obsolete_ui_json](https://github.com/writer/writer-framework/blob/dev/src/writer/wf_project.py#L164) that calls `write_files` without context

ramedina86:
Hey @UladzislauK-Writer and @mmikita95 , nice work!

I'd go for full on memory just to simplify things.

Please don't worry about `migrate_obsolete_ui_json` it's... in fact... obsolete itself :) so please delete all this logic

PR #854, opened 2025-05-02
fix: retry deepcopy of execution environment if failing due to concurrent modification

mmikita95:
A `RuntimeError` was intermittently raised during event handling:
<details>
<summary>Full traceback</summary>

```
A runtime exception was raised when processing event 'wf-click'.
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 34, in _get_executor
    yield current_app_process.executor
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 291, in execute_dag
    update_log("Executing...")
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 263, in update_log
    self._generate_run_log(tools, title, entry_type, msg=message, run_id=run_id)
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 206, in _generate_run_log
    "executionEnvironment": self._summarize_data_for_log(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 151, in _summarize_data_for_log
    data = copy.deepcopy(data)
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/copy.py", line 146, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/copy.py", line 271, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/copy.py", line 146, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/copy.py", line 230, in _deepcopy_dict
    for key, value in x.items():
RuntimeError: dictionary changed size during iteration

During handling of the above exception, another exception occurred:

  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/core.py", line 1765, in _handle_component_event
    self.blueprint_runner.execute_ui_trigger(target_id, ev.type, calling_arguments)
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 55, in execute_ui_trigger
    self.run_branch(trigger.id, None, execution_environment, "UI trigger execution")
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 114, in run_branch
    return self.execute_dag(nodes, execution_environment, title)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/site-packages/writer/blueprints.py", line 281, in execute_dag
    with self._get_executor() as executor:
  File "/Users/usr/.pyenv/versions/3.11.0/lib/python3.11/contextlib.py", line 190, in __exit__
    raise RuntimeError("generator didn't stop after throw()")
RuntimeError: generator didn't stop after throw()
```
</details>

The issue occurs during a call to `copy.deepcopy(data)` inside the `_summarize_data_for_log` method of `blueprints.py`, specifically when attempting to snapshot the `execution_environment` for logging. The root cause appears to be that the dictionary was being modified concurrently, leading to unsafe iteration during deep copying.

This PR introduces a safer snapshot mechanism to prevent crashes due to concurrent mutations:
- Added a `_safe_snapshot` method in BlueprintRunner:
    - Retries `copy.deepcopy` up to 3 times if a `RuntimeError` caused by concurrent dictionary mutation occurs.
    - Logs each retry attempt and a final error if snapshotting fails.
    - Falls back to returning an empty dict only as a last resort (with logging).
    - Replaced direct call to `copy.deepcopy` in `_summarize_data_for_log` with `_safe_snapshot`.

⸻

**Tradeoffs and considerations**

- This avoids using locks, which could introduce deadlocks or performance issues.
- The fallback to {} is considered safe in this context, but logging ensures visibility into snapshot failures.
- This is a pragmatic mitigation. A more structural long-term fix might involve:
    - Capturing the snapshot earlier (e.g. in `run_branch`) before any asynchronous mutations occur.
    - Using immutable state or dedicated logging-safe representations.

### DIFF
--- a/src/writer/blueprints.py
+++ b/src/writer/blueprints.py
@@ -147,8 +147,37 @@ class BlueprintRunner:
         hashed_id = hashlib.sha256(raw_id.encode()).hexdigest()[:24]
         return hashed_id
 
+    def _safe_snapshot(self, data):
+        """
+        Safely create a snapshot of the data to avoid issues with
+        concurrent modifications.
+        This is a workaround for the issue where the execution environment
+        is being modified concurrently, causing a RuntimeError during deepcopy.
+        """
+        retries = 3
+        for attempt in range(retries):
+            try:
+                return copy.deepcopy(data)
+            except RuntimeError as e:
+                if "dictionary changed size" in str(e):
+                    # Execution environment is being modified concurrently.
+                    # Retry the deepcopy operation.
+                    # Another option would be to use a lock, but that
+                    # could lead to deadlocks.
+                    logging.warning(
+                        "Retrying deepcopy due to concurrent "
+                        f"mutation (attempt {attempt+1})"
+                        )
+                    time.sleep(0.01)
+                else:
+                    raise  # Unrelated RuntimeError — re-raise
+        logging.error(
+            "Failed to create a stable snapshot of data after 3 attempts."
+            )
+        return {}  # As a last resort — still risky, but at least log it
+
     def _summarize_data_for_log(self, data):
-        data = copy.deepcopy(data)
+        data = self._safe_snapshot(data)
         MAX_ROWS = 100
         if isinstance(data, list):
             return [self._summarize_data_for_log(item) for item in data[:MAX_ROWS]]


mmikita95:
I was able to find two so far:
- [Call stack & trace modification](https://github.com/writer/writer-framework/blob/f475e3a1e158ec43a95de06a7745e0ee6b474831/src/writer/blueprints.py#L357-L358)
- [Logging](https://github.com/writer/writer-framework/blob/f475e3a1e158ec43a95de06a7745e0ee6b474831/src/writer/blocks/base_block.py#L190-L191) (and also for [response population](https://github.com/writer/writer-framework/blob/f475e3a1e158ec43a95de06a7745e0ee6b474831/src/writer/blocks/base_block.py#L205))

Any hints on other places to look?

ramedina86:
I think we can start with just the `deepcopy`, like you did (unless you know the conflict happened somewhere else). 

However, you shouldn't `with threading.Lock():` as this will create a new lock every time, the idea is to reuse the lock.

I asked ChatGPT to adapt the function and reuse the lock for the recursive calls (my guess is the conflict appears when two different children work at the same time)

------

import threading
import copy
import json

def _summarize_data_for_log(self, data, lock=None):
    if lock is None:
        lock = threading.Lock()

    with lock:
        data = copy.deepcopy(data)
        MAX_ROWS = 100

        if isinstance(data, list):
            return [self._summarize_data_for_log(item, lock=lock) for item in data[:MAX_ROWS]]

        if isinstance(data, dict):
            return {
                k: self._summarize_data_for_log(v, lock=lock)
                for i, (k, v) in enumerate(data.items())
                if i < MAX_ROWS
            }

        if isinstance(data, (str, int, float, bool, type(None))):
            return data

        try:
            return json.loads(json.dumps(data))
        except (TypeError, OverflowError):
            return f"Can't be displayed in the log. Value of type: {str(type(data))}."

mmikita95:
I'm not sure whether there's practicality in reusing the lock on recursive calls – they'd be working with deep-copied data, not the actual execution environment (and actually, this raises the question of whether we should deepcopy it too). Concurrent access happens only on first-level call, while we're still copying the live environment.

I'm thinking:
```
    def _summarize_data_for_log(self, data, is_recursive_call=False):
        if not is_recursive_call:
            with threading.Lock():
                # Prevent concurrent access to the data
                data = copy.deepcopy(data)
        MAX_ROWS = 100
        if isinstance(data, list):
            return [self._summarize_data_for_log(item, is_recursive_call=True) for item in data[:MAX_ROWS]]
        if isinstance(data, dict):
            return {
                k: self._summarize_data_for_log(v, is_recursive_call=True)
                for i, (k, v) in enumerate(data.items())
                if i < MAX_ROWS
            }
        if isinstance(data, list):
            return [self._summarize_data_for_log(item, is_recursive_call=True) for item in data]
        if isinstance(data, dict):
            return {k: self._summarize_data_for_log(v, is_recursive_call=True) for k, v in data.items()}
        if isinstance(data, (str, int, float, bool, type(None))):
            return data
        try:
            return json.loads(json.dumps(data))
        except (TypeError, OverflowError):
            return f"Can't be displayed in the log. Value of type: {str(type(data))}."
```

ramedina86:
You're right about it being pointless as it's the deep copied data... the issue is likely somewhere else. 

I think each "tool" will need its own lock.

`with threading.Lock():` is pointless though... defeats the purpose of locks.

Looks like we need to think harder about this.

ramedina86:
Hopefully we already took care of this with the `execution_environment_snapshot`

PR #858, opened 2025-05-07
feat: model selection field

mmikita95:
New `Model Id` field that enables a dropdown selection of Writer Palmyra LLMs for relevant blocks.

### DIFF
--- a/src/ui/src/builder/BuilderModelSelect.vue
+++ b/src/ui/src/builder/BuilderModelSelect.vue
@@ -1,0 +1,95 @@
+<script setup lang="ts">
+import injectionKeys from "@/injectionKeys";
+import {
+	computed,
+	defineAsyncComponent,
+	inject,
+	onMounted,
+	PropType,
+} from "vue";
+import { useListResources } from "@/composables/useListResources";
+import type { Option } from "@/wds/WdsSelect.vue";
+import BuilderAsyncLoader from "./BuilderAsyncLoader.vue";
+import type { WriterGraph } from "@/writerTypes";
+import WdsTextInput from "@/wds/WdsTextInput.vue";
+
+const WdsSelect = defineAsyncComponent({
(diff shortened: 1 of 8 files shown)


src/ui/src/builder/BuilderModelSelect.vue
```diff
@@ -0,0 +1,95 @@
...
+const currentValue = defineModel({
+	type: [String, Array] as PropType<string | string[]>,
+	required: true,
+});
+
+const {
+	load: loadGraphs,
+	data: graphs,
+	isLoading,
+} = useListResources<WriterGraph>(wf, "models");
```
madeindjs, line 36: You use the wrong type here (it works though because `WriterGraph` has `name` & `id`).

You should use introduce a new type `WriterModel` like this one

https://github.com/writer/writer-framework/blob/117920a0b05b13a17ad2a89138238f0d7c15a8e7/src/ui/src/writerTypes.ts#L212-L224

and use it here (with taking care of renaming vars names)
```suggestion
const {
	load: loadModel,
	data: models,
	isLoading,
} = useListResources<WriterModel>(wf, "models");
```

src/ui/src/builder/settings/BuilderFieldsWriterResourceId.vue
```diff
@@ -82,13 +94,23 @@ const ressourceUrl = computed(() => {
...
+	const def = wf.getComponentDefinition(component.value.type);
+	return def?.fields?.[props.fieldKey];
+});
+
 const selected = computed<string>({
-	get: () => component.value.content[props.fieldKey] ?? "",
+	get: () =>
+		component.value.content[props.fieldKey] ||
+		fieldDefinition.value.default ||
+		"",
```
madeindjs, line 113: Good catch ! 👏

PR #867, opened 2025-05-09
feat: run workflow block if at least one dependency is fulfilled - WF-298

raaymax:
https://writerai.atlassian.net/browse/WF-298

Screenshot 2025-05-09 at 13 27 25

### DIFF
--- a/src/writer/blueprints.py
+++ b/src/writer/blueprints.py
@@ -259,7 +259,7 @@ class BlueprintRunner:
         graph = {}
         in_degree = {node.id: 0 for node in nodes}
 
-        inputs = {}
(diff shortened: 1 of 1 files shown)


src/writer/blueprints.py
```diff
@@ -283,6 +307,8 @@ def update_log(message: str, entry_type="info"):
             while ready or futures:
                 while ready:
                     tool = ready.popleft()
+                    if not check_requirements(tool):
+                        break
```
ramedina86, line 311: I have some doubts about this... I'm spinning up a Scalingo environment to test.

PR #869, opened 2025-05-14
fix(writer): handle file operation on S3 - WF-386

madeindjs:


ramedina86:
Hey, as a reminder, we need a different approach because this won't work reliably with binary files

ramedina86:
I think in this context just treating everything as binary (text and binary) might work though

ramedina86:
Took care of this in a different PR - dir creation behavior still to be tested

PR #870, opened 2025-05-18
Fix dataclass defaults

wassemgtk:
## Summary
- prevent sharing state between `WfProjectContext` instances

## Testing
- `npm run cli:lint` *(fails: Cannot find implementation or library stub for module named "writer.ui")*
- `python3 -m ruff check src/writer/wf_project.py`


ramedina86:
Nice fix! I don't think it'll have an impact but definitely the right thing to do

PR #878, opened 2025-05-21
feat: API trigger

mmikita95:
Adds a new type of trigger that is supposed to make the blueprint API-compatible. Supports providing default payload.

Modifications to existing Jobs API:
- blueprints without API trigger are treated as unavailable via API
- authorization mechanism for job creation is removed – kept for job retrieval, but potentially can be removed as well
- `BlueprintRunner` is extended with a verifier that checks whether a blueprint is API-available

## Summary by CodeRabbit

- **New Features**
  - Introduced support for running blueprints via API triggers, allowing execution through dedicated API endpoints.
  - Added a new "API Trigger" block for blueprints, enabled via a feature flag.

- **Changes**
  - Updated job API endpoints to a new private path (`/private/api/job/...`).
  - Enhanced validation and error reporting when creating blueprint jobs via the API.
  - Removed the "Execute via API" option from the blueprint settings UI.
  - Refined blueprint event handling with explicit event types and structured payloads.
  - Filtered templates based on feature flags to control API trigger block visibility.
  - Changed job creation endpoint to stream status updates via server-sent events (SSE).
  - Simplified event handler logic by removing dynamic blueprint handler name resolution.

- **Bug Fixes**
  - Improved event handling and payload structure for blueprint execution with explicit event types.

- **Tests**
  - Updated tests to use new private API paths and removed authorization header requirements.
  - Revised tests to consume streaming SSE events for blueprint job status updates.


src/writer/core.py
```diff
@@ -1639,15 +1639,28 @@ def _handle_binding(self, event_type, target_component, instance_path, payload)
             return
         self.evaluator.set_state(binding["stateRef"], instance_path, payload)
 
+    def _confirm_blueprint_api_availability(self, blueprint_key: str):
+        def fn(payload, context, session):
+            if not self.blueprint_runner.is_blueprint_api_available(blueprint_key):
+                raise RuntimeError(
```
ramedina86: How does this break? Are we able to see it in the API call? I'd like for a 404 or another (400-499) relevant code to be added, with a message like this one.

mmikita95: >I'd like for a 404 or another (400-499) relevant code to be added, with a message like this one.

This happens for API call, but we have to await the event processing (`serve.py:418` to `serve.py:435`). 
This error specifically doesn't get to the `serve.py` – it only fires in app log. At `serve`-level, we need to await the result of the event handling to issue an HTTP exception, and we unfortunately don't have access to `AppProcess` / session / event handler & blueprint runner directly so that we could utilize the context here.

PR #902, opened 2025-05-29
Optimize source file tree building

wassemgtk:
## Summary
- avoid quadratic path lookups in `wf_project.build_source_files`

## Testing
- `pip install -e .` *(fails: Could not find a version that satisfies the requirement poetry-core)*

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability when handling file paths that contain duplicate directory names, ensuring correct directory checks in all cases.


ramedina86:
Thanks @wassemgtk and Codex ;)

PR #916, opened 2025-06-05
fix(ui): avoid rendering vault vars - WF-428

madeindjs:
image

## Summary by CodeRabbit

- **Bug Fixes**
  - Expressions starting with "vault" now display as masked with asterisks for enhanced security.


ramedina86:
I switched the mask for asterisks to make it clearer that we're hiding that on purpose, but other expressions will be interpolated without issues

PR #920, opened 2025-06-10
fix: error on non-alpha categories

UladzislauK-Writer:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved validation for category names to ensure they contain only English letters, digits, and underscores. Invalid category names will now trigger an error.
- **Tests**
	- Added a new test to confirm that invalid category names are correctly detected and result in an error.

### DIFF
--- a/src/writer/blocks/writerclassification.py
+++ b/src/writer/blocks/writerclassification.py
@@ -1,4 +1,5 @@
 import json
+import re
 
 from writer.abstract import register_abstract_template
 from writer.blocks.base_block import WriterBlock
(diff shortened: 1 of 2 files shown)


src/writer/blocks/writerclassification.py
```diff
@@ -27,7 +27,7 @@ def register(cls, type: str):
                             "name": "Categories",
                             "type": "Key-Value",
                             "default": "{}",
-                            "desc": "The keys should be the categories you want to classify the text in, for example 'valid' and 'invalid', and the values the criteria for each category.",
+                            "desc": "The keys should be the categories you want to classify the text in, for example 'valid' and 'invalid', and the values the criteria for each category. Keys should contain only letters of the English alphabet.",
```
ramedina86: Let's make it match Regex \w so 0-9, a-z, A-Z and underscore

src/ui/src/builder/useComponentActions.ts
```diff
@@ -854,6 +854,29 @@ export function useComponentActions(
 		wf.sendComponentUpdate();
 	}
 
+	/**
```
ramedina86: Let's leave this aside for now and rethink it later. Please remove `syncDynamicOuts` and all its mentions.

You can look into adding a watcher of dynamicOuts in BlueprintsNode but please do it in another PR

ramedina86 (changes requested):

PR #923, opened 2025-06-10
chore: update default agent to be more explicit on using Preview.

raaymax:
## Summary by CodeRabbit

- **Documentation**
  - Updated instructional text in the "Example customer review responder" section to clarify how users should preview and generate a response.


ramedina86:
@raaymax does this hit automatically? or we need to copy to S3?

raaymax:
Hopefully with my newest changes in agent-manager it should be propagated automatically

PR #925, opened 2025-06-11
fix: remove dynamic outs

UladzislauK-Writer:
## Summary by CodeRabbit

- **New Features**
  - Outputs that are no longer dynamically defined are now automatically removed from components, ensuring cleaner and more accurate output management.


ramedina86:
Neat PR, thanks! Also great suggestions by the rabbit, please implement them

UladzislauK-Writer:
Can't actually use `flush: "post"`. This leads to errors because we keep referencing deleted outs for a brief moment after rerender. It gets synced eventually because everything is correct after a refresh, but I guess it's happening too late

PR #947, opened 2025-06-18
feat: stop blueprint execution

raaymax:
## Summary by CodeRabbit

* **New Features**
  * Added the ability to stop (cancel) running blueprints from the UI, with updated controls and icons to reflect running and stopping states.
  * Introduced a new "cancelled" state for blueprint nodes, including distinct styling and icons to indicate cancelled executions.
  * Added a computed property to track the currently active blueprint run.

* **Improvements**
  * Enhanced logging and status tracking for blueprint runs, including better handling and display of cancelled executions.
  * Improved run lifecycle management with support for cancellation and stronger type and error handling guarantees.

* **Bug Fixes**
  * Improved robustness and error handling for blueprint execution and cancellation processes.

* **Tests**
  * Expanded and reorganized test coverage for blueprint execution, including new tests for cancellation, error handling, and complex graph scenarios.


src/ui/src/builder/builderManager.ts
```diff
@@ -406,6 +406,22 @@ export function generateBuilderManager() {
...
+			logEntries.find((entry) => {
+				return !!entry.blueprintExecution;
+			})?.blueprintExecution?.runId ?? null;
+		const isActive = !logEntries
+			.filter((entry) => {
+				return entry?.blueprintExecution?.runId === runId;
+			})
+			.find( (entry) => {
+				return entry.exit
+			});
```
madeindjs: It could be done in the same loop
```suggestion
			.filter((entry) => {
				return entry?.blueprintExecution?.runId === runId && entry.exit;
			});
```

src/ui/src/builder/builderManager.ts
```diff
@@ -406,6 +406,22 @@ export function generateBuilderManager() {
 		return state.value.logEntries;
 	};
 
+	const activeBlueprintRunId = computed(() => {
+		const logEntries = getLogEntries();
+		const runId =
+			logEntries.find((entry) => {
+				return !!entry.blueprintExecution;
+			})?.blueprintExecution?.runId ?? null;
```
madeindjs: If we don't have the `runId`, we could early return, no ?
```suggestion
		const runId =
			logEntries.find((entry) => {
				return !!entry.blueprintExecution;
			})?.blueprintExecution?.runId;
	    if (runId === undefined) return null;
```

src/ui/src/composables/useBlueprintRun.ts
```diff
@@ -89,7 +117,17 @@ export function useBlueprintRun(
...
 	}
 
-	return { isRunning: readonly(isRunning), run };
+	async function stop() {
+		const activeRunId = wfbm.activeBlueprintRunId.value;
+		if(!activeRunId) return;
+		try {
+			await stopBlueprintRun(wf, activeRunId);
+		} catch (err) {
+			console.error("Failed to stop blueprint run:", err);
```
madeindjs: Could you initialize `useLogger` and use it? It's just a proxy to `console` but it could be useful lat some point to centralize logs
```suggestion
			logger.error("Failed to stop blueprint run:", err);
```

src/ui/src/components/blueprints/base/BlueprintToolbar.vue
```diff
@@ -92,11 +99,12 @@ async function runBlueprint(componentId?: string) {
...
-			:disabled="isRunning"
-			@click="runBlueprint()"
+			@click="isRunning ? stopBlueprintRun() : runBlueprint()"
 		>
-			<i class="material-symbols-outlined">play_arrow</i>
-			{{ isRunning ? "Running..." : "Run blueprint" }}
+			<i class="material-symbols-outlined">{{
+				isRunning ? "stop" : "play_arrow"
+			}}</i>
+			{{ isRunning ? "Stop run" : "Run blueprint" }}
```
madeindjs, line 103: Don't you think we should use "Cancel" for the button ? 🤔 

```suggestion
			{{ isRunning ? "Cancel" : "Run blueprint" }}
```

src/ui/src/components/blueprints/base/BlueprintToolbar.vue
```diff
@@ -51,6 +55,10 @@ async function runBlueprint(componentId?: string) {
 	runBlueprintBtn.value?.toggleDropdown(false);
 	await handleRun(componentId);
 }
+
+async function stopBlueprintRun() {
+	await handleStop();
+}
```
madeindjs: You can avoid this alias and directly use `handleStop` https://github.com/writer/writer-framework/blob/b55c0325615105969e56746ae8f408aa97417297/src/ui/src/components/blueprints/base/BlueprintToolbar.vue#L102

madeindjs (approved):

src/ui/src/builder/builderManager.ts
raaymax: but is it more readable?
```
		const isActive = Boolean(
			logEntries
				.filter((entry) => {
					return entry?.blueprintExecution?.runId === runId
						&& !entry?.blueprintExecution?.exit;
				})?.[0]
		)
```

src/ui/src/components/blueprints/base/BlueprintToolbar.vue
raaymax: I found "Stop run" somewhere on designs and to be honest I don't have any opinion here 🤷

src/ui/src/composables/useBlueprintRun.ts
raaymax: Just noticed this error is handled by wf tracking, I don't think we need this log... removed

src/ui/src/builder/builderManager.ts
madeindjs: I didn't mean to inline everything, it was just to avoid an extra loop on the array

PR #951, opened 2025-06-19
chore: multi-selection for graphs in Ask graph question block - WF-564

mmikita95:
Introduces a new type of field for graph ID multi-selection & enables it for "Ask graph question" block

## Summary by CodeRabbit

- **New Features**
	- Added support for selecting multiple graph IDs in relevant fields within the builder settings interface.
- **Improvements**
	- Enhanced the resource selection component to allow multi-selection, improving flexibility for users managing graph-related configurations.


madeindjs (approved):
I think it could have been possible to reuse `BuilderFieldsWriterResourceId` with a new param

src/ui/src/builder/settings/BuilderFieldsWriterResourceId.vue
```diff
@@ -83,18 +89,21 @@ const linkTooltip = computed(() => {
 });
 
 const ressourceUrl = computed(() => {
-	if (!selected.value) return;
+	const value = Array.isArray(selected.value)
```
madeindjs: I guess we can't display the resource URL if user selected multiple items. I think we should early return if `selected.value` is an array

madeindjs (approved):

PR #954, opened 2025-06-19
feat: add autogen button to components

UladzislauK-Writer:
## Summary by CodeRabbit

- **New Features**
	- Added an "Autogenerate" button to the sidebar toolkit, allowing users to trigger autogeneration functionality when working with blueprints.
- **Style**
	- Updated panel and sidebar layouts for improved spacing and appearance, including dynamic panel switcher height and enhanced footer visuals.
	- Adjusted sizing and layout of sidebar elements to be more responsive using CSS variables and flexbox.
- **Refactor**
	- Improved internal state management for the autogeneration feature, ensuring consistent visibility and control across components.

### DIFF
--- a/src/ui/src/builder/BuilderApp.vue
+++ b/src/ui/src/builder/BuilderApp.vue
@@ -115,6 +115,8 @@ import {
 	inject,
 	onMounted,
 	onUnmounted,
+	provide,
+	ref,
 	watch,
 	useTemplateRef,
 } from "vue";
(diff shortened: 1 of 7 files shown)


src/ui/src/builder/BuilderApp.vue
```diff
@@ -136,6 +138,12 @@ import BuilderInstanceTracker from "./BuilderInstanceTracker.vue";
 import BuilderCollaborationTracker from "./BuilderCollaborationTracker.vue";
 import BaseNote from "@/components/core/base/BaseNote.vue";
 
+const isAutogenShown = ref(false);
+provide(injectionKeys.isAutogenShown, isAutogenShown);
+provide(injectionKeys.showAutogen, () => {
+	isAutogenShown.value = true;
+});
```
madeindjs: I don't think you need two `provide` here. You could simply have

```suggestion
provide(injectionKeys.isAutogenShown, ref(false));
```

And then open the autogen like this

```ts
const isAutogenShown = inject(injectionKeys.isAutogenShown, ref(false));

function showAutogen() {
  isAutogenShown.value = true;
}
```

Also, I think it could be better to rename to `isAutogenModalShown`

PR #958, opened 2025-06-20
fix: constraint file rename

UladzislauK-Writer:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved file path validation when renaming files to prevent use of invalid or reserved paths.
  - Added safeguards to block file operations within restricted directories, enhancing security and preventing access to internal files.

- **New Features**
  - Users now receive clear error messages if they attempt to rename files using invalid or restricted paths.

### DIFF
--- a/src/ui/src/builder/panels/BuilderCodePanel.vue
+++ b/src/ui/src/builder/panels/BuilderCodePanel.vue
@@ -268,8 +268,21 @@ async function handleSave() {
 
 	isDisabled.value = true;
 
+	const pathSegments = isRenaming.value
+		? filename.value.split("/")
+		: undefined;
+	if (pathSegments) {
+		const errorMessage = checkRenamePath(pathSegments);
+		if (errorMessage) {
+			return pushToast({
+				type: "error",
+				message: errorMessage,
+			});
+		}
+	}
+
 	try {
-		await save(isRenaming.value ? filename.value.split("/") : undefined);
+		await save(pathSegments);
 		pushToast({ type: "success", message: "The file was saved" });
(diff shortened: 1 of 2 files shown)


src/ui/src/builder/panels/BuilderCodePanel.vue
```diff
@@ -268,8 +268,20 @@ async function handleSave() {
 
 	isDisabled.value = true;
 
+	let pathSegments: string[] | undefined = undefined;
+	if (isRenaming.value) {
+		pathSegments = filename.value.split("/");
```
madeindjs: I guess you can improve the readability by avoiding the re-assignation 

```suggestion
	const pathSegments = isRenaming.value ? filename.value.split("/") : undefined;
	if (pathSegments) {
```

src/ui/src/builder/panels/BuilderCodePanel.vue
```diff
@@ -280,6 +292,22 @@ async function handleSave() {
...
+
+	if (pathSegments[0] === ".wf") {
+		return ".wf is reserved for Writer Framework";
+	}
+
+	if (pathSegments.some((segment) => segment === "")) {
+		return "Invalid path";
+	}
+
+	return null;
```
madeindjs: Optional, but I think you can shorten it

```suggestion
	if (pathSegments[0] === "") return "Path should not have a leading slash";
	if (pathSegments[0] === ".wf") return ".wf is reserved for Writer Framework";
	if (pathSegments.some((segment) => segment === "")) return "Invalid path";
```

madeindjs (approved):
Minor comments but LGTM

PR #965, opened 2025-06-23
fix: evaluate full match

UladzislauK-Writer:
## Summary by CodeRabbit

- **Tests**
	- Added new test to verify full-match template replacements in field evaluation.
	- Extended existing tests to cover additional escaped and Unicode character scenarios.
- **Chores**
	- Updated component blueprints with additional classification components referencing full-match variables.
- **Bug Fixes**
	- Improved JSON string escaping in template replacements for more accurate output.


src/writer/evaluator.py
```diff
@@ -55,7 +55,13 @@ def decode_json(text):
             raise ValueError(f'Component with id "{component_id}" not found.')
 
         field_value = component.content.get(field_key) or default_field_value
-        replaced = None
+
+        if as_json:
+            # For full match remove quotes from the string because it was serialized
```
ramedina86: What does this mean?

src/writer/evaluator.py
```diff
@@ -55,7 +55,13 @@ def decode_json(text):
             raise ValueError(f'Component with id "{component_id}" not found.')
 
         field_value = component.content.get(field_key) or default_field_value
-        replaced = None
```
UladzislauK-Writer, line 58: We have assignments to replaced both in `if` and `else` branches later. So this one is redundant

src/writer/evaluator.py
UladzislauK-Writer: Basically, what I posted in Slack: that the field value is always JSON-serialized and we end up with  `"\"@{full_match}\""` and for full matches we need to remove outer quotes. You found this unclear, so I should rephrase

ramedina86 (approved):
Looking good, feel free to merge when ready

PR #988, opened 2025-07-01
fix: row jumping for dataframe

UladzislauK-Writer:
## Summary by CodeRabbit

* **New Features**
  * Textarea cells now automatically adjust their height to fit content when text wrapping is enabled.
  * Improved text wrapping and overflow handling for both editable and non-editable text cells.

* **Style**
  * Enhanced styles for multiline text wrapping and overflow behavior in text cells.

* **Refactor**
  * Updated component structure for number cells by adding an additional wrapper for improved layout consistency.

### DIFF
--- a/src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellNumber.vue
+++ b/src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellNumber.vue
@@ -20,14 +20,16 @@ function onChange(newValue: number) {
 </script>
 
 <template>
-	<WdsNumberInput
-		v-if="editable"
-		class="CoreDataframeCellNumber--input"
-		:model-value="Number(value)"
-		@focusout="onChange($event)"
-	/>
-	<div v-else class="CoreDataframeCellNumber--text">
-		{{ value }}
+	<div>
+		<WdsNumberInput
+			v-if="editable"
+			class="CoreDataframeCellNumber--input"
+			:model-value="Number(value)"
+			@focusout="onChange($event)"
+		/>
+		<div v-else class="CoreDataframeCellNumber--text">
+			{{ value }}
+		</div>
 	</div>
 </template>
 
--- a/src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellNumber.vue
+++ b/src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellNumber.vue
@@ -20,16 +20,15 @@ function onChange(newValue: number) {
 </script>
 
 <template>
-	<div>
+	<div v-if="editable">
 		<WdsNumberInput
-			v-if="editable"
 			class="CoreDataframeCellNumber--input"
 			:model-value="Number(value)"
 			@focusout="onChange($event)"
 		/>
-		<div v-else class="CoreDataframeCellNumber--text">
-			{{ value }}
-		</div>
+	</div>
+	<div v-else class="CoreDataframeCellNumber--text">
+		{{ value }}
 	</div>
 </template>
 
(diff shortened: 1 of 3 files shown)


src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellText.vue
```diff
@@ -45,6 +77,11 @@ function stopEditing() {
 .CoreDataframeCellText--text {
 	width: 100%;
 	font-size: inherit;
+	overflow: hidden;
+	white-space: pre-wrap;
+	word-wrap: break-word;
+	overflow-wrap: break-word;
+	word-break: break-word;
```
UladzislauK-Writer, line 77: `overflow: hidden` only for `noWrap` caused the issue on the screenshot

image

Everything else I just kind of threw in there, both old-style and modern-style to just make sure that in both `p` and `textarea` we have the same number of lines. Because line-breaks would mess with it. And very long words would stretch the `textarea` and make other long lines not wrap properly

src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellNumber.vue
```diff
@@ -20,14 +20,16 @@ function onChange(newValue: number) {
...
+		<WdsNumberInput
+			v-if="editable"
+			class="CoreDataframeCellNumber--input"
+			:model-value="Number(value)"
+			@focusout="onChange($event)"
+		/>
+		<div v-else class="CoreDataframeCellNumber--text">
+			{{ value }}
+		</div>
 	</div>
```
madeindjs: I'm not sure that you need the extra `div` in read-only mode

```suggestion
	<div v-if="editable">
		<WdsNumberInput
			class="CoreDataframeCellNumber--input"
			:model-value="Number(value)"
			@focusout="onChange($event)"
		/>
	</div>
	<div v-else class="CoreDataframeCellNumber--text">
		{{ value }}
	</div>
```

src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellText.vue
```diff
@@ -21,6 +21,38 @@ function stopEditing() {
...
+});
+
+watch(
+	() => props.editable || props.value,
+	() => {
+		if (props.wrapText && props.editable) {
+			nextTick(() => autoResizeFromRef());
+		}
+	},
+);
```
madeindjs: 1. you can use `immediate: true` to execute when mounted
2. you use the callback props to refactor a bit
```suggestion
watch(
	() => props.editable && props.value,
	(shouldResize) => {
		if (shouldResize) {
			nextTick(() => autoResizeFromRef());
		}
	},
	{ immediate: true }
);
```

src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellText.vue
```diff
@@ -21,6 +21,38 @@ function stopEditing() {
...
+
+	const el = textarea.value.ref as HTMLTextAreaElement;
+
+	const style = getComputedStyle(el);
+	if (!style) return;
+
+	el.style.height = "auto";
+
+	const borderTop = parseFloat(style.borderTopWidth || "0");
+	const borderBottom = parseFloat(style.borderBottomWidth || "0");
```
madeindjs: 1. I guess it's better to avoid `NaN` this way
2. we can use integer for px
```suggestion
	const borderTop = parseInt(style.borderTopWidth) || 0;
	const borderBottom = parseInt(style.borderBottomWidth) || 0;
```

src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellText.vue
```diff
@@ -21,6 +21,38 @@ function stopEditing() {
...
+
+function autoResizeFromRef() {
+	if (!textarea.value?.ref) return;
+
+	const el = textarea.value.ref as HTMLTextAreaElement;
+
+	const style = getComputedStyle(el);
+	if (!style) return;
+
+	el.style.height = "auto";
```
madeindjs: Why do we need to assign twice ?

https://github.com/writer/writer-framework/blob/cce465f67db192c434598751d9baa91d32d555ce/src/ui/src/components/core/content/CoreDataframe/CoreDataframeCellText.vue#L39

PR #996, opened 2025-07-02
feat(ui): more working space in Text Area and Code modals - AB-99

VladKrasny:
- Remove obsolete `variant` prop from `BuilderTemplateInput` component
- Improve `WdsModal` layout to stretch scrollable content
- Apply 50vh height to TextArea and Code builder fields when in extended mode

## Summary by CodeRabbit

* **New Features**
  * Introduced a "half-screen" display option for the embedded code editor, providing a new visual variant with a distinct height.
  * Added dynamic resizing for textarea fields, allowing them to expand to half the screen height when in expanded mode.

* **Enhancements**
  * Improved modal layout with flexible sizing and vertical stacking for better usability and appearance.
  * Updated textarea controls to support dynamic row count and custom styling, enhancing the editing experience.


src/ui/src/builder/settings/BuilderTemplateInput.vue
```diff
@@ -101,10 +104,10 @@ const props = defineProps({
 	inputId: { type: String, required: false, default: undefined },
 	value: { type: String, required: false, default: undefined },
 	multiline: { type: Boolean, required: false },
-	variant: {
-		type: String as PropType<"code" | "text">,
+	rows: {
+		type: [String, Number] as PropType<number | "half-screen">,
```
madeindjs: It's a bit weird that this component can accept `"half-screen"` as the number of rows. It's a sign that you're mixing responsibilities. It should stay generic.

It's better to keep only `Number` here.
```suggestion
		type: Number,
```

Then you need to handle this logic in the parent

https://github.com/writer/writer-framework/blob/c016bf8261d44838d3ac4fe0e8f3fbbe1ce1ca7b/src/ui/src/builder/settings/BuilderFieldsText.vue#L24

like doing 

```
:rows="isExpanded ? undefined : 3"
```

Maybe you'll need to remove with `default: 3` and refactor a bit

src/ui/src/builder/settings/BuilderTemplateInput.vue
```diff
@@ -101,10 +104,10 @@ const props = defineProps({
 	inputId: { type: String, required: false, default: undefined },
 	value: { type: String, required: false, default: undefined },
 	multiline: { type: Boolean, required: false },
-	variant: {
-		type: String as PropType<"code" | "text">,
+	rows: {
```
madeindjs: since `rows` only makes sens only when it's `multiline`, I would prefix it like
```suggestion
	multilineRows: {
```

src/ui/src/builder/BuilderEmbeddedCodeEditor.vue
```diff
@@ -108,6 +114,10 @@ onUnmounted(() => {
 	min-height: 300px;
 }
 
+.BuilderEmbeddedCodeEditor--half-screen {
```
madeindjs: Nitpick: Usually, we use `camelCase` for modifiers.
```suggestion
.BuilderEmbeddedCodeEditor--halfScreen {
```

madeindjs (changes requested):

src/ui/src/builder/settings/BuilderTemplateInput.vue
VladKrasny: Thanks Alex, I see your point and it makes total sense to me.

Let me explain my idea behind the change. I used "half-screen" more as a declarative way to control height, similar to how rows works, just more semantic.

If we move that logic to the parent, we'll probably need to expose styles or classes for the inner input/textarea, since right now all styling applies to the root.

src/ui/src/builder/settings/BuilderTemplateInput.vue
```diff
@@ -95,9 +97,17 @@ const emit = defineEmits(["input", "update:value"]);
...
 	inputId: { type: String, required: false, default: undefined },
+	inputStyle: {
+		type: Object as PropType<CSSProperties>,
+		default: () => ({}),
+	},
 	componentId: { type: String, required: false, default: undefined },
 	value: { type: String, required: false, default: undefined },
 	multiline: { type: Boolean, required: false },
+	multilineRows: {
+		type: [Number] as PropType<number>,
```
madeindjs:
```suggestion
		type: Number,
```

madeindjs (approved):

PR #1010, opened 2025-07-11
feat(ui): migrate from Material to Lucide icons - AB-90

VladKrasny:
## Summary by CodeRabbit

* **New Features**
  * Introduced a unified icon component for consistent icon rendering across the UI.
  * Added new custom and Lucide icon support, enabling a broader and more modern icon set.

* **Refactor**
  * Replaced all inline and font-based icons with the new unified icon component throughout the application.
  * Updated icon names and visual styles for improved clarity and consistency.
  * Refined button and icon spacing for better alignment and appearance.

* **Bug Fixes**
  * Fixed a typo in button link properties (disabled state now works as expected).

* **Style**
  * Improved icon and button alignment with updated CSS for a more polished look.

* **Chores**
  * Updated dependencies to support new icon functionality.

* **Tests**
  * Adjusted tests to validate icon rendering using the new component.


src/ui/src/wds/WdsTag.vue
```diff
@@ -31,7 +32,7 @@ defineEmits({
 			class="WdsTag__close"
 			@click.stop="$emit('close')"
 		>
-			<i class="material-symbols-outlined">close</i>
+			<WdsIcon name="x" />
```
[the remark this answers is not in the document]
VladKrasny: `x` is a valid icon that is available in the lib - https://lucide.dev/icons/x

src/ui/src/builder/BuilderBlueprintState.vue
```diff
@@ -42,16 +43,18 @@ const result = computed(() => {
...
 		<slot v-if="!result" name="unknown"></slot>
 		<slot v-else-if="result === 'success'" name="success">
-			<span
-				class="BuilderBlueprintState__status BuilderBlueprintState__status--success material-symbols-outlined"
-				>check</span
+			<div
+				class="BuilderBlueprintState__status BuilderBlueprintState__status--success"
 			>
+				<WdsIcon name="check" />
+			</div>
```
madeindjs, line 50: Do we need an extra `<div>`? Can't it be 
```suggestion
			<WdsIcon
				class="BuilderBlueprintState__status BuilderBlueprintState__status--success"
				name="check"
			/>
```

It applies to others.

src/ui/src/wds/WdsIcons/LucideIcon.vue
```diff
@@ -0,0 +1,48 @@
...
+	<span :style="{ display: 'contents' }" v-html="rawHtml" />
+</template>
+
+<script setup lang="ts">
+import { nextTick, onMounted, ref, useAttrs, watch } from "vue";
+import { createIcons, icons } from "lucide";
+
+defineOptions({
+	inheritAttrs: false,
+});
```
madeindjs, line 13: Why do you need this? It prevents us from doing  `<WdsIcon class="foo" />`

src/ui/src/wds/WdsIcons/LucideIcon.vue
```diff
@@ -0,0 +1,48 @@
...
+watch(
+	() => props.name,
+	(newName) => {
+		renderIcon(newName);
+	},
+);
+
+onMounted(() => {
+	renderIcon(props.name);
+});
```
madeindjs: You can use `immediate` instead of mixing `watch` / `onMounted`. Also,It can be shortened.

```suggestion
watch(toRef(props, 'name'),  renderIcon, { immediate: true });
```

src/ui/src/wds/WdsIcons/LucideIcon.vue
```diff
@@ -0,0 +1,48 @@
...
+const props = defineProps({
+	name: { type: String, required: true },
+});
+
+const attrs = useAttrs();
+
+const rawHtml = ref("");
+
+function renderIcon(name: string) {
+	rawHtml.value = `<i data-lucide="${name}"></i>`;
```
madeindjs: Do you really need `v-html` ?
```suggestion
	<span :style="{ display: 'contents' }">
	  <i :data-lucide="name"></i>
	</span>
</template>

<script setup lang="ts">
import { nextTick, onMounted, ref, useAttrs, watch } from "vue";
import { createIcons, icons } from "lucide";

defineOptions({
	inheritAttrs: false,
});

const props = defineProps({
	name: { type: String, required: true },
});

const attrs = useAttrs();

const rawHtml = ref("");

function renderIcon(name: string) {
	rawHtml.value = ``;
```

src/ui/src/wds/WdsIcons/LucideIcon.vue
```diff
@@ -0,0 +1,48 @@
...
+
+	nextTick(() => {
+		createIcons({
+			icons,
+			attrs: {
+				width: "1em",
+				height: "1em",
```
madeindjs, line 31: I think it's better to have `width` & `height` as props. Also, we don't use `em`

```suggestion
const props = defineProps({
	name: { type: String, required: true },
	size: { type: String, required: false, default: '12px' },
});

const rawHtml = ref("");

function renderIcon(name: string) {
	rawHtml.value = `<i data-lucide="${name}"></i>`;

	nextTick(() => {
		createIcons({
			icons,
			attrs: {
				width: props.size,
				height: props.size,
```

src/ui/src/wds/WdsIcons/CustomIcon.vue
```diff
@@ -0,0 +1,68 @@
...
+		</svg>
+	</template>
+</template>
+
+<script lang="ts">
+const iconNames = new Set([
+	"wds-blocks",
+	"wds-function",
+	"wds-blueprints",
+] as const);
```
madeindjs, line 65: Out of curiosity, why do we need the `as const` ?

src/ui/src/wds/WdsIcons/CustomIcon.vue
```diff
@@ -0,0 +1,68 @@
...
+</script>
+
+<script setup lang="ts">
+import { PropType } from "vue";
+
+defineProps({
+	name: { type: String as PropType<CustomIconName>, required: true },
+});
+
+const size = "1em";
```
madeindjs, line 81:
```suggestion
	name: { type: String as PropType<CustomIconName>, required: true },
	size: { type: String, required: false, default: '12px' },
});
```

src/ui/src/wds/WdsIcons/CustomIcon.vue
```diff
@@ -0,0 +1,68 @@
...
+	"wds-blocks",
+	"wds-function",
+	"wds-blueprints",
+] as const);
+
+type CustomIconName = typeof iconNames extends Set<infer T> ? T : never;
+
+export function isCustomIconName(name: string): name is CustomIconName {
+	return iconNames.has(name as CustomIconName);
+}
```
madeindjs, line 71: TypeScript is awesome. 😄

madeindjs (commented):
Well done, I left small comments

src/ui/src/wds/WdsIcons/LucideIcon.vue
VladKrasny: Great catch, I will apply it!

Originally the logic relied on the DOM element being available, which is why I used onMounted. But since that changed, I didn't update the implementation.

src/ui/src/wds/WdsIcons/LucideIcon.vue
VladKrasny: Yes, in this case we do need `v-html`

The reason is that some components pass icon names dynamically (e.g. `chevron-down` vs `chevron-up`). And Lucide `createIcons` function replaces the `<i>` element with actual SVG **only once**.

When Vue re-renders, it doesn't get updated, since Lucide doesn't re-process already replaced elements (no reactive logic).

Using `v-html` lets us re-inject a fresh `<i>` node on every change and trigger `createIcons` cleanly on nextTick.

src/ui/src/wds/WdsIcons/LucideIcon.vue
VladKrasny: It prevents class and style from being applied to the outer `<span>` wrapper (which uses display: contents), since that element shouldn't affect layout or styling. Instead I'm manually forwarding those attributes to the `<i>` tag that gets replaced by Lucide. So the real SVG element will get all the attributes you are passing.

src/ui/src/wds/WdsIcons/CustomIcon.vue
VladKrasny: Yeah, good question 😁

I used `as const` so that TypeScript sees each value as a specific string literal instead of just `string` type.
Without it, the set would be `Set<string>` and the type `CustomIconName` would just be `string`.  But with it we get an exact union like `Set<"wds-blocks" | "wds-function" | "wds-blueprints">`.

you can see more here - https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

src/ui/src/wds/WdsIcons/LucideIcon.vue
VladKrasny: Yeah, that one is tricky. I used em intentionally to match the behavior of Material icons, since they are essentially icon fonts that scale with the font-size of the parent. You can check that `material-symbols-outlined` class uses font-size: 100%, so I wanted to preserve that dynamic scaling for SVG icons too.

But I'm totally fine switching to a fixed size prop (like px) if that's the direction we want to take. But we would need to make sure it doesn't break any cases where icon size was previously inherited via font-size.

src/ui/src/builder/BuilderBlueprintState.vue
VladKrasny: In that particular case we do need the extra `<div>` as it acts as a styled circle with background and fixed size, and the icon is centered inside it. If we removed the wrapper, the SVG itself would stretch to fill the space (unlike icon fonts), which would break the intended layout and styling. This setup preserves the original look we had with icon fonts.

Here is a visual example. The icon on the left has no wrapper, and the one on the right does have the wrapper:
Screenshot 2025-07-14 at 12 44 04

src/ui/src/wds/WdsIcons/LucideIcon.vue
VladKrasny: We might also need to add `flush: 'post'` to ensure the DOM is ready before calling `createIcons`, since it relies on actual elements being in place - https://vuejs.org/guide/essentials/watchers.html#post-watchers

src/ui/src/wds/WdsIcons/CustomIcon.vue
madeindjs: Ha , I see. Nice move!

src/ui/src/wds/WdsIcons/LucideIcon.vue
madeindjs: Good point, so I guess we can keep your implementation :+1:

madeindjs (approved):
You'll need to rebase and update the snapshot of `SharedJsonViewer.spec.ts` using `npm run test.watch` but it looks good to me :muscle:

PR #1024, opened 2025-07-22
use orjson for websockets data

UladzislauK-Writer:
## Summary by CodeRabbit

* **New Features**
  * Improved WebSocket communication by sending JSON data in a binary format for enhanced performance and compatibility.
  * Added support for reading and parsing JSON data from Blob objects in WebSocket messages.

* **Bug Fixes**
  * Enhanced error handling for WebSocket message parsing and Blob data reception.

* **Chores**
  * Added the orjson dependency to support efficient binary serialization of JSON data.

* **Tests**
  * Updated tests to handle binary WebSocket messages and ensure correct JSON parsing from received bytes.


src/ui/src/core/index.ts
```diff
@@ -310,6 +308,41 @@ export function generateCore() {
...
+						"Error in readBlobAsArrayBuffer callback: ",
+						error,
+					);
+				}
+			};
+			reader.onerror = (error) => {
+				logger.error("Error reading Blob:", error);
+			};
+			reader.readAsArrayBuffer(blob);
+		}
```
madeindjs: Since the core file is already big, could you move this as utils function in `src/utils/blob.ts` ? Also, I think it's more reusable if we promisify

```ts
// src/utils/blob.ts
export function readBlobAsArrayBuffer(
  blob: Blob,
): Promise<string | ArrayBuffer | null> {
  return new Promise((res, rej) => {
    const reader = new FileReader();
    reader.onload = () => {
      try {
        res(reader.result);
      } catch (error) {
        rej(error);
      }
    };
    reader.onerror = (error) => {
      rej(error);
    };
    reader.readAsArrayBuffer(blob);
  });
}

export async function readBlobAsArrayBufferJson<T extends unknown>(
  blob: Blob,
): Promise<T | undefined> {
  const buffer = await readBlobAsArrayBuffer(blob);

  if (!buffer) return undefined;
  if (typeof buffer === "string") return JSON.parse(buffer);

  const decoder = new TextDecoder("utf-8");
  return JSON.parse(decoder.decode(buffer));
}
```

src/ui/src/core/index.ts
```diff
@@ -310,6 +308,41 @@ export function generateCore() {
...
+
+		webSocket.onmessage = (wsEvent) => {
+			if (wsEvent.data instanceof Blob) {
+				readBlobAsArrayBuffer(wsEvent.data, (buffer) => {
+					const decoder = new TextDecoder("utf-8");
+					const message = JSON.parse(decoder.decode(buffer));
+					processMessage(message);
+				});
```
madeindjs: You need to wrap it in a `try/catch`, no ?

PR #1025, opened 2025-07-22
fix: switch to emit instead of accessing property of child - AB-180

mmikita95:
Accessing property enabled a race condition, where app data could be possibly not yet available upon retrieval. Switching to event listening allows to mitigate the issue and ensure that data is always available.

## Summary by CodeRabbit

* **Refactor**
  * Updated application selection to focus on single application selection instead of multiple graphs.
  * Improved event-driven handling for application selection updates.

* **New Features**
  * Application selection now emits an event with the selected application's details for easier integration.

* **Bug Fixes**
  * Ensured selected application data is accurately reflected and updated in related fields.

* **Other**
  * Extended application data type to optionally include input parameters.

### DIFF
--- a/src/ui/src/builder/BuilderApplicationSelect.vue
+++ b/src/ui/src/builder/BuilderApplicationSelect.vue
@@ -6,6 +6,7 @@ import {
 	inject,
 	onMounted,
 	PropType,
+	watch,
 } from "vue";
 import { useListResources } from "@/composables/useListResources";
 import type { Option } from "@/wds/WdsSelect.vue";
@@ -20,6 +21,10 @@ const WdsSelect = defineAsyncComponent({
 
 const wf = inject(injectionKeys.core);
 
+const emit = defineEmits<{
+	(event: "selected-data", data: WriterApplication | undefined): void;
+}>();
+
 defineProps({
 	enableMultiSelection: { type: Boolean, required: false },
 });
@@ -30,12 +35,12 @@ const currentValue = defineModel({
 });
 
 const {
-	load: loadGraphs,
-	data: graphs,
+	load: loadApps,
+	data: apps,
 	isLoading,
 } = useListResources<WriterApplication>(wf, "applications");
 
-onMounted(loadGraphs);
+onMounted(loadApps);
 
 function getAppType(app: WriterApplication) {
 	switch (app.type) {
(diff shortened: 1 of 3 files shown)


src/ui/src/builder/settings/BuilderFieldsWriterResourceId.vue
```diff
@@ -149,20 +151,25 @@ const selected = computed<string | string[]>({
...
-	if (props.resourceType !== "application" || !newAppId) return;
-
-	const appData = selectorEl.value?.selectedData;
-	const inputsList = appData?.inputs;
-
-	if (!inputsList || typeof inputsList !== "object") return;
+const selectedData = ref<WriterApplication | undefined>();
+
+function onSelectedData(appData: WriterApplication | undefined) {
+	selectedData.value = appData; // for readability
```
madeindjs, line 157: I don't understand why we need to hold this value in the memory. Can't it be
```suggestion
function onSelectedData(appData: WriterApplication | undefined) {
```

src/ui/src/builder/BuilderApplicationSelect.vue
```diff
@@ -76,9 +81,14 @@ const currentValueStr = computed<string>({
...
-		? graphs.value.find((g) => g.id === currentValue.value)
-		: graphs.value.filter((g) => currentValue.value.includes(g.id));
+	// Application select should never support multi-selection
+	if (typeof currentValue.value !== "string") return undefined;
+
+	return apps.value.find((g) => g.id === currentValue.value);
+});
+
+watch(selectedData, (data) => {
+	emit("selected-data", data);
```
madeindjs: We'll have this event dispatched every time we load the component, and so we'll call `setContentValue(component.value.id, "appInputs"` everytime we open the settings.

I wonder if it's not better to do something like

```vue
<WdsSelect  @update:model-value="onSelect" />
```
```ts
function onSelect() {
  emit("selected-data", apps.value.find((g) => g.id === currentValue.value));
}
```
WDYT ?

mmikita95: Makes sense - and now that you've mentioned it, I probably need to make sure we don't override user-defined inputs. Thank you, will fix!

PR #1045, opened 2025-07-29
fix(ui): graceful handling for refused websocket connection; error display w/ reason catching

mmikita95:
- Replaced the previous error handling with a div that shows a message explaining why initialization failed, removing the loading spinner before displaying it; div now potentially can be extended with a class name and additional design
- Updated the WebSocket startup logic to cleanly remove temporary listeners and provide an explicit error when the socket closes before establishing a connection
- Added error logging and improved the thrown error message in the retry loop of initSession to handle cases where no response is received

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during initialization to prevent runtime errors and provide clearer error messages if initialization fails.
  * Enhanced WebSocket connection handling to provide explicit error messages when connections close prematurely.
  * Updated error display to present user-friendly, accessible messages with retry instructions and a reload button, replacing the previous page overwrite and removing the loading spinner.

### DIFF
--- a/src/ui/src/core/index.ts
+++ b/src/ui/src/core/index.ts
@@ -129,7 +129,7 @@ export function generateCore() {
 			} catch {
 				if (attempt >= maxRetries - 1) {
 					throw new Error(
-						`Failed to acquire initialization data. Server responded with ${response.status} status.`,
+						`Failed to acquire initialization data. Server responded with ${response?.status ?? "no response"} status.`,
 					);
 				}
 				await new Promise((r) => setTimeout(r, RECONNECT_DELAY_MS * (attempt + 1)));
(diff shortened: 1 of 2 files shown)


src/ui/src/core/index.ts
```diff
@@ -362,8 +362,25 @@ export function generateCore() {
...
+					),
+				);
+			}
+			function cleanup() {
+				webSocket.removeEventListener("open", handleOpen);
+				webSocket.removeEventListener("close", handleClose);
+			}
+
+			webSocket.addEventListener("open", handleOpen);
+			webSocket.addEventListener("close", handleClose);
```
madeindjs:
```suggestion
			webSocket.addEventListener("open", () => resolve(), { once: true });
			webSocket.addEventListener("close", () => {
				reject(new Error("WebSocket connection closed before establishing.")
			}), { once: true });
```

madeindjs (approved):

PR #1054, opened 2025-08-01
feat(ui) - define graphs IDs using variables in state - AB-91 

VladKrasny:
## Summary by CodeRabbit

* **New Features**
  * Added a toggle button for switching between binding and non-binding modes in field settings.
  * Introduced new composables for managing binding mode and field values reactively.
  * Enhanced support for resource ID fields with improved multi-selection and default value handling.

* **Refactor**
  * Updated UI components to centralize and streamline binding logic and state management.
  * Consolidated and simplified rendering logic for resource ID field types.

* **Style**
  * Improved tooltips and icons for binding toggle buttons to enhance user clarity.

### DIFF
--- a/src/ui/src/builder/settings/composables/useBindingMode.ts
+++ b/src/ui/src/builder/settings/composables/useBindingMode.ts
@@ -1,4 +1,4 @@
-import { ref, DeepReadonly, Ref } from "vue";
+import { ref, DeepReadonly, Ref, readonly } from "vue";
 
 function checkIsBindingFormat(val: unknown): boolean {
 	return typeof val === "string" && val.startsWith("@{") && val.endsWith("}");
@@ -37,7 +37,7 @@ export function useBindingMode({ fieldValue, setFieldValue }: Params) {
 	}
 
 	return {
-		isBindingMode,
+		isBindingMode: readonly(isBindingMode),
 		toggleBindingMode,
 	};
 }
--- a/src/ui/src/builder/settings/composables/useBindingMode.ts
+++ b/src/ui/src/builder/settings/composables/useBindingMode.ts
@@ -1,0 +1,43 @@
+import { ref, DeepReadonly, Ref } from "vue";
+
+function checkIsBindingFormat(val: unknown): boolean {
+	return typeof val === "string" && val.startsWith("@{") && val.endsWith("}");
+}
+
+type Params = {
+	fieldValue: DeepReadonly<Ref<string>>;
+	setFieldValue: (value: string) => void;
+};
+
+export function useBindingMode({ fieldValue, setFieldValue }: Params) {
+	const isBindingMode = ref<boolean>(checkIsBindingFormat(fieldValue.value));
+
+	const lastValues = ref<{
+		binding: string;
+		noneBinding: string;
+	}>({
+		binding: checkIsBindingFormat(fieldValue.value) ? fieldValue.value : "",
+		noneBinding: isBindingMode.value ? "" : fieldValue.value,
+	});
+
+	function toggleBindingMode() {
+		if (isBindingMode.value) {
+			if (checkIsBindingFormat(fieldValue.value)) {
+				lastValues.value.binding = fieldValue.value;
+			}
+
+			setFieldValue(lastValues.value.noneBinding);
+			isBindingMode.value = false;
+		} else {
+			lastValues.value.noneBinding = fieldValue.value;
+
+			setFieldValue(lastValues.value.binding);
+			isBindingMode.value = true;
+		}
+	}
+
+	return {
+		isBindingMode,
+		toggleBindingMode,
+	};
+}
--- a/src/ui/src/builder/settings/composables/useBindingMode.ts
+++ b/src/ui/src/builder/settings/composables/useBindingMode.ts
@@ -1,37 +1,40 @@
-import { ref, DeepReadonly, Ref, readonly } from "vue";
+import { ref, Ref, readonly } from "vue";
 
 function checkIsBindingFormat(val: unknown): boolean {
 	return typeof val === "string" && val.startsWith("@{") && val.endsWith("}");
 }
 
 type Params = {
-	fieldValue: DeepReadonly<Ref<string>>;
-	setFieldValue: (value: string) => void;
+	fieldViewModel: Ref<string>;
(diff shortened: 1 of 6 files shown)


src/ui/src/builder/useComponentFieldViewModel.ts
```diff
@@ -0,0 +1,45 @@
...
+
+type Params = {
+	componentId: DeepReadonly<Ref<string>>;
+	fieldKey: DeepReadonly<Ref<string>>;
+	defaultValue?: DeepReadonly<Ref<string>>;
+};
+
+export function useComponentFieldViewModel(params: Params) {
+	const wf = inject(injectionKeys.core);
+	const ssbm = inject(injectionKeys.builderManager);
```
madeindjs: It's always better to "inject" dependencies; I'll make your composable more testable and more reusable.

```suggestion
export function useComponentFieldViewModel(
	params: Params,
	wf = inject(injectionKeys.core),
	ssbm = inject(injectionKeys.builderManager),
) {
```

src/ui/src/builder/useComponentFieldViewModel.ts
```diff
@@ -0,0 +1,45 @@
...
+		set: (value: string) => {
+			setFieldValue(value);
+		},
+	});
+
+	return {
+		setFieldValue,
+		fieldValue,
+		fieldViewModel,
+	};
```
madeindjs: Why do do you need to expose getter/setting ? Could you only return `fieldViewModel`
```suggestion
	return fieldViewModel;
```

src/ui/src/builder/settings/composables/useBindingMode.ts
```diff
@@ -0,0 +1,43 @@
...
+		} else {
+			lastValues.value.noneBinding = fieldValue.value;
+
+			setFieldValue(lastValues.value.binding);
+			isBindingMode.value = true;
+		}
+	}
+
+	return {
+		isBindingMode,
```
madeindjs: It's better to use `readonly` to be sure it can't be toggled mannually
```suggestion
		isBindingMode: readonly(isBindingMode),
```

src/ui/src/builder/useComponentFieldViewModel.ts
```diff
@@ -0,0 +1,45 @@
+import { computed, inject, Ref, DeepReadonly } from "vue";
+import injectionKeys from "@/injectionKeys";
+import { useComponentActions } from "./useComponentActions";
+
+type Params = {
+	componentId: DeepReadonly<Ref<string>>;
+	fieldKey: DeepReadonly<Ref<string>>;
+	defaultValue?: DeepReadonly<Ref<string>>;
+};
```
madeindjs: I think you want to use `MaybeRef` for `fieldKey`
```suggestion
type Params = {
	componentId: DeepReadonly<Ref<string>>;
	fieldKey: MaybeRef<string>;
	defaultValue?: DeepReadonly<Ref<string>>;
};
```

It makes your function accept `string | Ref<string>`, so you don't have to wrap your string like this

https://github.com/writer/writer-framework/blob/f9470c683698a0b2f3bb8f181179a9e432d3252f/src/ui/src/builder/settings/BuilderFieldsWriterResourceId.vue#L99

Then you just need to use `unref()` when you want to use the value

VladKrasny: great suggestion!
I haven't thought about that.

src/ui/src/builder/useComponentFieldViewModel.ts
VladKrasny: I exposed getter/setter separately because in some components it can be useful to apply transformations before assignment, without having to wrap everything in another computed. It gives more flexibility to the parent to decide how to use the view model either directly or with custom logic.

src/ui/src/builder/useComponentFieldViewModel.ts
VladKrasny: > make your composable more testable and more reusable

good pint, I fully agree with that!
but I've not seen that way of setting inject() as default parameters and I could not find that example in Vue documentation or their gihtub. 

If needed for testing, I can still mock the injects by passing them explicitly:

```ts
type Dependencies = {
	wf?: Core;
	ssbm?: BuilderManager;
};

export function useComponentFieldViewModel(
	params: Params,
	dependencies?: Dependencies,
) {
	const wf = dependencies?.wf ?? inject(injectionKeys.core);
	const ssbm = dependencies?.ssbm ?? inject(injectionKeys.builderManager);
```

src/ui/src/builder/useComponentFieldViewModel.ts
madeindjs: I still think it's better to only expose the model, It looks redundant to have both. When you need to wrap the setter, it doesn't help, no ?

```ts
// actual
const { setFieldValue } = useComponentFieldViewModel(/* ... */)

function onSetFieldValue(/* ... */) {
  setFieldValue(foo)
  // other logic
}

// model
const model = useComponentFieldViewModel(/* ... */)

function setFieldValue(/* ... */) {
  model.value = foo
  // other logic
}

```

src/ui/src/builder/useComponentFieldViewModel.ts
madeindjs: I think the default value to `inject` is supported by Vue.js.

But thinking about it again, I think the standard way in the application is like this

https://github.com/writer/writer-framework/blob/21bb1b15c665e84fcd0689e020441d65f123b071/src/ui/src/core/useNotesManager.ts#L34-L37
https://github.com/writer/writer-framework/blob/21bb1b15c665e84fcd0689e020441d65f123b071/src/ui/src/composables/useBlueprintRun.ts#L103-L107

So I think it should be

```ts
export function useComponentFieldViewModel(
 	wf: ReturnType<typeof generateCore>, 
 	wfbm: ReturnType<typeof generateBuilderManager>, 
	params: Params,
) {
```

src/ui/src/builder/useComponentFieldViewModel.ts
VladKrasny: good point, I will change that to expose only the model

madeindjs (approved):

PR #1059, opened 2025-08-05
feat(ui): unify builder settings fields handling - AB-210

VladKrasny:
## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal state management across multiple builder field components by adopting a unified field view model, resulting in more consistent and maintainable field value handling.
  * Updated input bindings and event handlers to use a new reactive approach, reducing manual state synchronization and direct content mutation.
  * Adjusted default value handling for checkbox, object, and text fields to support more flexible defaults.
  * Improved internal logic for various field types (text, color, padding, shadow, width, etc.) for a more reliable and predictable user experience.
  * Centralized alias and blueprint key management using the new field view model composable for simplified editing and selection.
  * Enhanced tools and notes components to utilize reactive field view models for data updates, improving responsiveness and consistency.
  * Refined event extraction logic in settings to safely handle missing component definitions.

* **Style**
  * Updated layout of the Writer Resource ID field to use a flexbox layout for improved alignment.

* **Chores**
  * Minor updates to internal type definitions to standardize parameters for field view models.
  * Simplified JSON parsing and default value fallbacks in evaluation utilities for more consistent behavior.


src/ui/src/builder/settings/BuilderFieldsAlign.vue
```diff
@@ -24,7 +24,7 @@
 			<BuilderTemplateInput
 				v-if="mode == 'css'"
 				ref="freehandInputEl"
-				:value="component.content[fieldKey]"
+				:value="fieldViewModel"
```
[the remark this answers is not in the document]
VladKrasny: In Vue templates refs are automatically unwrapped, so there is no need to explicitly use .value here. The current binding works as intended.

src/ui/src/builder/settings/BuilderFieldsAlign.vue
```diff
@@ -155,8 +150,13 @@ const props = defineProps({
 	error: { type: String, required: false, default: undefined },
 });
 
-const { componentId, fieldKey, direction } = toRefs(props);
-const component = computed(() => wf.getComponentById(componentId.value));
+const { direction } = toRefs(props);
+
+const fieldViewModel = useComponentFieldViewModel({
+	componentId: toRef(props, "componentId"),
+	fieldKey: toRef(props, "fieldKey"),
```
madeindjs:
```suggestion
const { direction, componentId, fieldKey } = toRefs(props);

const fieldViewModel = useComponentFieldViewModel({
	componentId,
	fieldKey,
```

src/ui/src/builder/settings/BuilderFieldsKeyValue.vue
```diff
@@ -81,33 +81,30 @@ const props = defineProps({
 	error: { type: String, required: false, default: undefined },
 });
 
+const fieldViewModel = useComponentFieldViewModel({
+	componentId: toRef(props, "componentId"),
+	fieldKey: toRef(props, "fieldKey"),
+	defaultValue: "",
```
madeindjs: You are not using the default value here. Some KeyValue fields has a default value set

https://github.com/writer/writer-framework/blob/aa186380e32ba3ccc32253d7a84428e0c5de4fe1/src/ui/src/components/core/input/CoreCheckboxInput.vue#L74-L80

https://github.com/writer/writer-framework/blob/aa186380e32ba3ccc32253d7a84428e0c5de4fe1/src/ui/src/components/core/input/CoreSelectInput.vue#L64-L70

This is also true for some other types .

You need to be sure your pass `:default-value` everywhere in `BuilderSettingsProperties.vue`.

Or you can also compute the default value in the hook:

```ts
type Params = {
	componentId: MaybeRef<string>;
	fieldKey: MaybeRef<string>;
};

export function useComponentFieldViewModel(
	params: Params,
	dependencies?: Dependencies,
) {
	const wf = dependencies?.wf ?? inject(injectionKeys.core);
	const ssbm = dependencies?.ssbm ?? inject(injectionKeys.builderManager);

	const component = computed(() => wf.getComponentById(params.componentId.value));

	const def = computed(() => wf.getComponentDefinition(component.value?.type));

	const defaultValue = computed(() => def.value?.[params.field.value]?.default);
```

madeindjs (changes requested):
Nice refactoring. I think I spotted an issue with the default value on some fields.

src/ui/src/builder/settings/BuilderFieldsKeyValue.vue
VladKrasny: As usual you are right, but I did not pass the default value intentionally to preserve the current behavior (with no defaults).

However, I agree this looks more like a bug in the existing logic, so the reasonable action would be to include default value propagation as part of this change

src/ui/src/renderer/useEvaluator.ts
```diff
@@ -195,16 +195,15 @@ export function useEvaluator(wf: Core, secretsManager?: SecretsManager) {
...
 			}
 			if (typeof evaluated !== "string") return evaluated;
-			let parsedValue: any;
+
 			try {
-				parsedValue = JSON.parse(evaluated);
+				return JSON.parse(evaluated);
 			} catch {
-				return JSON.parse(defaultValue ?? null);
+				return JSON.parse(defaultValue || null);
```
madeindjs, line 205: well seen

madeindjs (approved):

PR #1073, opened 2025-08-11
fix: prevent fallback value from overwriting cleared input by tracking dirty state - AB-220

VladKrasny:
## Summary by CodeRabbit

- Bug Fixes
  - Edited component fields no longer get overwritten by fallback values; user-entered text now reliably persists.
  - Prevents unexpected resets when switching fields or updating content, improving editing consistency.
  - Current input is prioritized over fallback content for string fields for a more predictable experience.

- Tests
  - Added unit tests for field view model initialization, reactive defaults, and persisting edits.

- Chores
  - Cleaned up public type exports and simplified logger return type for a clearer API surface.

### DIFF
--- a/src/ui/src/builder/useComponentFieldViewModel.spec.ts
+++ b/src/ui/src/builder/useComponentFieldViewModel.spec.ts
@@ -100,7 +100,15 @@ describe(useComponentFieldViewModel.name, () => {
 		expect(
 			mockCore.core.getComponentById(componentId).content[fieldKey],
 		).toBe("new value");
-
 		expect(vm.value).toBe("new value");
+
+		vm.value = "";
+
+		await nextTick();
+
+		expect(
+			mockCore.core.getComponentById(componentId).content[fieldKey],
+		).toBe("");
+		expect(vm.value).toBe("");
 	});
 });
--- a/src/ui/src/builder/useComponentFieldViewModel.spec.ts
+++ b/src/ui/src/builder/useComponentFieldViewModel.spec.ts
@@ -1,51 +1,55 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
+import { describe, it, expect } from "vitest";
+import { ref, nextTick } from "vue";
+import { buildMockComponent, buildMockCore } from "@/tests/mocks";
+import { generateBuilderManager } from "./builderManager";
+import {
+	useComponentFieldViewModel,
+	Dependencies,
+} from "./useComponentFieldViewModel";
+import { FieldType, WriterComponentDefinition } from "@/writerTypes";
 
-import { describe, it, expect, vi } from "vitest";
-import { ref, reactive, nextTick } from "vue";
-import { useComponentFieldViewModel } from "./useComponentFieldViewModel";
-
-type Component = { id: string; type: string; content: Record<string, string> };
-type Defs = Record<string, { fields?: Record<string, { default?: string }> }>;
-
-function mockDependencies(
-	initial: Record<string, Component>,
-	defs: Defs = {},
-): { components: Record<string, Component>; wf: any; ssbm: any } {
-	const components = reactive(initial);
-
-	function getComponentById(id: string) {
-		return components[id];
-	}
-
-	function getComponentDefinition(type?: string) {
-		return type ? defs[type] : undefined;
-	}
-
-	return {
-		components,
-		wf: {
-			getComponentById,
-			getComponentDefinition,
-			sendComponentUpdate: vi.fn(),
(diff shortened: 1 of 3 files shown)


src/ui/src/builder/useComponentFieldViewModel.ts
```diff
@@ -72,16 +72,23 @@ export function useComponentFieldViewModel(
...
-			fallbackValue.value
-		);
+		const val = component.value?.content?.[toValue(params.fieldKey)];
+
+		if (isDirty.value && typeof val === "string") {
+			return val;
+		}
+
+		return val || fallbackValue.value;
 	});
```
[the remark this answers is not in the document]
VladKrasny: this logic is correct, the dirty param ensures that cleared string remains empty

madeindjs (commented):
Could you add a test that ensures the behavior is correct ?

src/ui/src/builder/useComponentFieldViewModel.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+describe("useComponentFieldViewModel", () => {
+	const componentId = "TestComponent";
+	const fieldKey = "TestField";
+
+	function setupMockDependencies(componentId: string, fieldKey: string) {
+		return mockDependencies(
+			{ [componentId]: { id: componentId, type: "Text", content: {} } },
+			{ Text: { fields: { [fieldKey]: { default: "default test" } } } },
```
madeindjs: You can use `buildMockComponent` which creates a valid button by default.

```suggestion
			{ [componentId]: buildMockComponent({ id: componentId, content: {} }) },
```

src/ui/src/builder/useComponentFieldViewModel.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+			sendComponentUpdate: vi.fn(),
+		},
+		ssbm: {
+			openMutationTransaction: vi.fn(),
+			registerPreMutation: vi.fn(),
+			registerPostMutation: vi.fn(),
+			closeMutationTransaction: vi.fn(),
+		},
+	};
+}
```
madeindjs: We already have `buildMockCore` that takes care of mocking API call

```suggestion
function mockDependencies(
	initial: Record<string, Component>,
): { components: Record<string, Component>; wf: any; ssbm: any } {
	const components = reactive(initial);
	
	const mockCore = buildMockCore();
	for (const comp of Object.values(initial)) mockCore.core.addComponent(comp);

	return {
		components,
		wf: mockCore.core,
		ssbm: {
			openMutationTransaction: vi.fn(),
			registerPreMutation: vi.fn(),
			registerPostMutation: vi.fn(),
			closeMutationTransaction: vi.fn(),
		},
	};
}
```

In the same vein, it could be the right time to introduce `buildMockBuilderManager` that can be reused.

src/ui/src/builder/useComponentFieldViewModel.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+		vm.value = "new value";
+
+		await nextTick();
+
+		expect(dependencies.components[componentId].content[fieldKey]).toBe(
+			"new value",
+		);
+
+		expect(vm.value).toBe("new value");
```
madeindjs: If you use `mockCore`, you could try to reusue the s
```suggestion
		expect(mockCore.core.getComponentById(componentId)!.content[fieldKey]).toBe(
			"new value",
		);

		expect(vm.value).toBe("new value");
```

src/ui/src/builder/useComponentFieldViewModel.spec.ts
VladKrasny: there is `generateBuilderManager` is already in place

src/ui/src/builder/useComponentFieldViewModel.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+					},
+				},
+			};
+		}
+
+		return {
+			wf: {
+				...mockCore.core,
+				getComponentDefinition,
+			},
```
madeindjs: It's better to rely on `vi.spyOn`

```suggestion
		vi.spyOn(mockCore.core, "getComponentDefinition").mockReturnValue({
			name: "",
			description: "",
			fields: {
				[fieldKey]: {
					name: "",
					type: FieldType.Text,
					default: "default test",
				},
			},
		});

		return {
			wf: mockCore.core,
```

madeindjs (approved):

PR #1100, opened 2025-08-22
feat(CoreFileInput): support drag and drop files as well as click to open finder - AB-27 

VladKrasny:
## Summary by CodeRabbit

- New Features
  - Drag-and-drop file upload area with type restrictions, optional total-size limit, and multi-file support.
  - New file item UI showing name, size, upload status, download action, and size-exceeded cues.
- Improvements
  - Centralized file management with encoding, progress/abort support, clearer uploading/error states, and ability to re-upload same files.
  - Simplified upload interactions and consistent size validation messages.
- Tests
  - Updated end-to-end selector for file input.
- Chores
  - Added runtime dependency for file-type acceptance.
- Removed
  - Legacy builder drop-zone component replaced by the new drop zone.


src/ui/src/composables/useFilesEncoder/useFilesEncoder.ts
```diff
@@ -0,0 +1,126 @@
...
+function convertFileToUiFile(file: File): UiFile {
+	const id = window.crypto?.randomUUID() ?? `file-${fallbackId++}`;
+
+	return {
+		id,
+		file,
+		name: file.name,
+		size: file.size,
+	};
+}
```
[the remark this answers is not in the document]
VladKrasny: no SSR support is needed

src/ui/src/composables/useFilesEncoder/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./useFilesEncoder";
```
madeindjs: Barrel files are a bad practice; it's better to avoid them ([good article](https://dev.to/tassiofront/barrel-files-and-why-you-should-stop-using-them-now-bc4))

src/ui/src/composables/useFilesEncoder/useFilesEncoder.ts
```diff
@@ -0,0 +1,114 @@
...
+	type: string;
+	data: string;
+};
+
+export type UseFilesEncoderParams = {
+	multiple: MaybeRef<boolean>;
+};
+
+export function useFilesEncoder({ multiple }: UseFilesEncoderParams) {
+	const uiFiles = ref<UiFile[]>([]);
```
madeindjs: It's better to use `shallowRef`, it's save a bit of memory

```suggestion
	const uiFiles = shallowRef<UiFile[]>([]);
```

src/ui/src/composables/useFilesEncoder/useFilesEncoder.ts
```diff
@@ -0,0 +1,114 @@
...
+		uiFiles.value = [];
+	}
+
+	async function encodeFiles(): Promise<{
+		encodedFiles: EncodedFile[];
+		rejectedFiles: Error[];
+	}> {
+		const settledResults = await Promise.allSettled(
+			uiFiles.value.map(async ({ file }) => {
+				const encodedFile = await encodeFileAsDataURL(file);
```
madeindjs: You don't use Signal here, no? You can use `useAbortController` and pass the abort.

src/ui/src/composables/useFilesEncoder/useFilesEncoder.ts
```diff
@@ -0,0 +1,114 @@
...
+			? uiFiles.value.concat(convertedFiles)
+			: convertedFiles.slice(0, 1);
+	}
+
+	function removeFile(id: string) {
+		const index = uiFiles.value.findIndex((uiFile) => uiFile.id === id);
+
+		if (index !== -1) {
+			uiFiles.value.splice(index, 1);
+		}
```
madeindjs: If you use `shallowRef`
```suggestion
		uiFiles.value = uiFiles.value.filter((uiFile) => uiFile.id !== id);
```

src/ui/src/composables/useFileTypeAccept.ts
madeindjs, line 1: I'm wondering if we should use https://www.npmjs.com/package/attr-accept instead . I think we are no covering all cases (like `*.pdf`)

https://github.com/user-attachments/assets/9bedf239-0117-4663-8707-20e0170d76c5

src/ui/src/composables/useFileTypeAccept.ts
```diff
@@ -0,0 +1,85 @@
...
+		if (normalizedAcceptedFileTypes.value.length === 0) {
+			// no restrictions, all files are accepted
+			return true;
+		}
+
+		const fileType = (file.type || "").toLowerCase();
+		const fileName = (file.name || "").toLowerCase();
+
+		const dotIndex = fileName.lastIndexOf(".");
+		const fileExtension = dotIndex > -1 ? fileName.slice(dotIndex) : "";
```
madeindjs: What about `*.tar.gz` ? 😈

src/ui/src/components/shared/SharedFile.vue
```diff
@@ -0,0 +1,109 @@
...
+		<span class="SharedFile__name" :title="name">
+			{{ name }}
+		</span>
+		<div v-if="areActionsShown" class="SharedFile__actions">
+			<WdsTag v-if="isSizeShown" variant="normal" :text="formattedSize" />
+			<WdsButton
+				v-if="isDownloadButtonShown"
+				variant="tertiary"
+				size="smallIcon"
+				data-writer-tooltip="Download File"
```
madeindjs, line 16: I think the tooltip should be on left 🤔 
image

src/ui/src/components/shared/SharedDropZone.vue
madeindjs, line 1: Could you reuse this component in `BuilderPanel` too? 

https://github.com/writer/writer-framework/blob/b44a949d4f3a550c14052f5f513275849ce72a10/src/ui/src/builder/panels/BuilderPanel.vue#L36-L36

src/ui/src/components/core/input/CoreFileInput.vue
```diff
@@ -118,102 +98,75 @@ export default {
...
-}
 
-const fileLabel = computed(() =>
-	allowMultipleFilesFlag.value ? "files" : "file",
-);
-const triggerText = computed(() => `Browse ${fileLabel.value}`);
-const emptyMessage = computed(() => `No ${fileLabel.value} selected`);
+	setTimeout(() => {
+		window.URL.revokeObjectURL(blobURL);
+	}, 200);
```
madeindjs, line 181: Why do we need `setTimeout` here? Could it be replaced by `await nextTick()` (which is a bit more predictable)?

VladKrasny: setTimeout is used as we need a macrotask to avoid revoking the URL before the download begins. It cannot be replaced with nextTick as it schedules a microtask.

src/ui/src/composables/useFileTypeAccept.ts
VladKrasny: good catch!

src/ui/src/composables/useFileTypeAccept.ts
VladKrasny: Good call on `attr-accept` I will replace the custom checker with the lib!

But `*.pdf` is not a valid accept token. The HTML spec supports MIME types like `application/pdf`, type wildcards like `application/*`, and extensions prefixed with a dot like `.pdf`.
Because of that, neither the native `<input accept="*.pdf">` not `attr-accept` lib will match `*.pdf`. So the lib will not fix that particular pattern. 

I can add an additional check in the normalization step and transform `*.pdf -> .pdf`

src/ui/src/components/shared/SharedDropZone.vue
VladKrasny: great suggestion!

Screenshot 2025-08-25 at 16 05 29

src/ui/src/composables/useFileTypeAccept.ts
```diff
@@ -0,0 +1,53 @@
...
+		return [...result];
+	});
+
+	const acceptAttr = computed<string | undefined>(
+		() => normalizedAcceptedFileTypes.value.join(",") || undefined,
+	);
+
+	function checkIsFileAccepted(file: File): boolean {
+		return accept(file, normalizedAcceptedFileTypes.value);
+	}
```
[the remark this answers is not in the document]
VladKrasny: It's fine to leave it as is. The library performs an empty array check internally and returns `true` meaning accept all

https://github.com/react-dropzone/attr-accept/blob/main/src/index.js#L17

src/ui/src/builder/panels/BuilderPanel.vue
```diff
@@ -33,7 +33,13 @@
...
 					@dragleave="handleDragEvent($event, false)"
 				>
-					<BuilderDropFileZone v-if="showDropingFilesZone" />
+					<SharedDropZone
+						v-if="showDropingFilesZone"
+						multiple
+						label="Drop files here"
+						:restrictions="`Files can be up to ${maxFileSizeLimit}`"
+						@drop="handleDropFiles"
+					/>
```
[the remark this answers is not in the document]
VladKrasny: File handling occurs at the parent level, so the suggestion will not be applied

madeindjs (approved):

PR #1126, opened 2025-09-08
feat(ui): improve offline experience - AB-308

madeindjs:
Use a job queue that will store component update socket calls when the backend is not running

## Summary by CodeRabbit

* **New Features**
  * Serialized promise queue for ordered frontend operations.
  * Deep plain-object equality check to detect real state changes.

* **Bug Fixes**
  * Waits for healthy sync state before sending messages.
  * Prevents unintended component state mutations during async updates by restoring unchanged state.

* **Tests**
  * Unit tests covering queue sequencing and error-handling behavior.

### DIFF
--- a/src/ui/src/composables/usePromiseQueue.ts
+++ b/src/ui/src/composables/usePromiseQueue.ts
@@ -15,7 +15,7 @@ export function usePromiseQueue(
 	const jobs: Job[] = [];
 	let isRunning = false;
 
-	async function add<T>(func: () => Promise<T>): Promise<T> {
+	function add<T>(func: () => Promise<T>): Promise<T> {
 		// eslint-disable-next-line no-async-promise-executor
 		return new Promise(async (res, rej) => {
 			async function callback() {
--- a/src/ui/src/composables/usePromiseQueue.ts
+++ b/src/ui/src/composables/usePromiseQueue.ts
@@ -1,0 +1,62 @@
+interface Job {
+	id: number;
+	func: () => Promise<void>;
+}
+type Hook = (job: Job) => Promise<void>;
+
+export function usePromiseQueue(
+	hooks: {
+		onJobExecuted?: Hook;
+		onJobResolved?: Hook;
+		onJobRejected?: Hook;
+	} = {},
+) {
+	let nextId = 0;
+	const jobs: Job[] = [];
+	let isRunning = false;
+
+	async function add<T>(func: () => Promise<T>): Promise<T> {
+		// eslint-disable-next-line no-async-promise-executor
+		return new Promise(async (res, rej) => {
+			async function callback() {
+				try {
+					const data = await func();
+					res(data);
+				} catch (e) {
+					rej(e);
+					throw e;
+				}
+			}
+
+			const job = { func: callback, id: incrementNextId() };
+			jobs.push(job);
+			await run();
+		});
+	}
+
+	async function run() {
+		if (isRunning) return;
+		isRunning = true;
+
+		while (jobs.length > 0) {
+			const job = jobs.shift();
+			if (!job) continue;
+			try {
+				await job.func();
+				if (hooks.onJobResolved) await hooks.onJobResolved(job);
+			} catch {
+				if (hooks.onJobRejected) await hooks.onJobRejected(job);
+			} finally {
+				if (hooks.onJobExecuted) await hooks.onJobExecuted(job);
+			}
+		}
+
+		isRunning = false;
+	}
+
+	function incrementNextId() {
+		return nextId++;
+	}
+
+	return { add };
+}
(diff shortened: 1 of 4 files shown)


src/ui/src/composables/usePromiseQueue.ts
```diff
@@ -0,0 +1,62 @@
...
+					rej(e);
+					throw e;
+				}
+			}
+
+			const job = { func: callback, id: incrementNextId() };
+			jobs.push(job);
+			await run();
+		});
+	}
```
bybash:
```suggestion
async function add<T>(func: () => Promise<T>): Promise<T> {
    return new Promise((res, rej) => {  // Not async!
        const job = { 
            func: async () => {
                try {
                    res(await func());
                } catch (e) {
                    rej(e);
                    // Don't re-throw
                }
            }, 
            id: incrementNextId() 
        };
        jobs.push(job);
        run();  // Trigger without await
    });
}
```

bybash: I think we don't need to rethrow error on exception and no need to wait for run.

src/ui/src/composables/usePromiseQueue.ts
```diff
@@ -0,0 +1,62 @@
...
+				await job.func();
+				if (hooks.onJobResolved) await hooks.onJobResolved(job);
+			} catch {
+				if (hooks.onJobRejected) await hooks.onJobRejected(job);
+			} finally {
+				if (hooks.onJobExecuted) await hooks.onJobExecuted(job);
+			}
+		}
+
+		isRunning = false;
```
bybash, line 54: If all jobs finish and isRunning becomes false, but a job was added during the final job's execution, that job might be stuck in the queue until another job is added to trigger a new run().
So this might be helpful:

```
isRunning = false;
// check if jobs were added while we were finishing
    if (jobs.length > 0) {
        run(); // Process any jobs added during the race window
    }
```

src/ui/src/core/index.ts
```diff
@@ -709,6 +725,30 @@ export function generateCore() {
...
+		if (syncHealth.value === value) return;
+
+		// eslint-disable-next-line no-async-promise-executor
+		return new Promise<void>(async (res) => {
+			while (syncHealth.value !== value) {
+				await new Promise((r) => setTimeout(r, 500));
+			}
+			return res();
+		});
+	}
```
bybash, line 739: What if it never returns true and loops forever? 
And you can take look at code rabbits comment too.

src/ui/src/core/index.ts
```diff
@@ -709,6 +725,30 @@ export function generateCore() {
...
+	async function sendFrontendMessageQueue(
+		type: string,
+		payload: object | (() => Promise<object>),
+		callback?: Function,
+		track = false,
+	) {
+		return queue.add(async () => {
+			await waitSyncHealthToBe("connected");
+			return sendFrontendMessage(type, payload, callback, track);
+		});
```
bybash, line 750: It seems you don't need to return anything.
```suggestion
		queue.add(async () => {
			await waitSyncHealthToBe("connected");
			sendFrontendMessage(type, payload, callback, track);
		});
```

src/ui/src/core/index.ts
```diff
@@ -794,8 +835,15 @@ export function generateCore() {
 					return reject("Couldn't connect to the server.");
 				}
 				resolve();
+				if (!isPlainObjectDeepEqual(components.value, copy)) {
```
bybash, line 839: Are we sure they don't have arrays as a prop?

bybash (changes requested):
Nice one, added some notes 👍

src/ui/src/core/index.ts
madeindjs: `components` is always a dictionary of component `Record<string, Component>`

src/ui/src/core/index.ts
madeindjs: It is not equivalent.

`queue.add` is adding the async func to the queue and running it. I omit to `return sendFrontendMessage()`, it won't wait and return before the message is sent.

src/ui/src/core/index.ts
madeindjs: That's the purpose of this PR: wait for the status to be back to "online" and hold back frontend messages until then.

Even if we're waiting for a status that never comes, I think this is still fine. The loop is basic and should consume a negligible amount of CPU/memory.

src/ui/src/composables/usePromiseQueue.ts
madeindjs: I don't see how it can happen.

If we add a new job, it'll be detected by the `while (jobs.length > 0)` or the `add()` method will start again the runner with `run()`

bybash: So we sure that jobs array will be updated inside that function. Ok, my question was a bit react oriented maybe.

src/ui/src/composables/usePromiseQueue.ts
```diff
@@ -0,0 +1,62 @@
...
+	}
```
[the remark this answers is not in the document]
bybash: We can add this, wdyt @madeindjs ?

PR #1145, opened 2025-10-24
Add block to access Writer Key-Value storage

UladzislauK-Writer:
## Summary by CodeRabbit

* **New Features**
  * Added a workflow block for key‑value storage operations (save/get/delete) with outcome reporting and error handling.
  * Added a client for remote key‑value storage and exposed a singleton for app‑wide use.
  * Vault secret retrieval now reads from the centralized key‑value storage backend.

* **Chores**
  * Added logging entries for KV storage and vault to improve observability.

### DIFF
--- a/src/writer/vault.py
+++ b/src/writer/vault.py
@@ -7,86 +7,13 @@ configuration and error handling.
 """
 
 import logging
-import os
-from functools import partial
-from typing import Any, Dict, Literal, Optional, Protocol
+from typing import Dict, Optional
 
 import requests
 
-logger = logging.getLogger("kv_storage")
+from writer.keyvalue_storage import writer_kv_storage
 
-
-class _WrappedRequestFunc(Protocol):
-    def __call__(
-        self,
-        headers: Dict[str, str],
-        timeout: int,
-    ) -> requests.Response: ...
-
-
-class KeyValueStorage:
-    def __init__(self) -> None:
-        base_url = os.getenv("WRITER_BASE_URL")
-        self.api_key = os.getenv("WRITER_API_KEY")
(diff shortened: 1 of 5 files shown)


src/writer/vault.py
```diff
@@ -8,27 +8,29 @@
...
-    def get_secrets(self) -> Dict:
-        """Get cached secrets, fetching from vault if not already loaded."""
-        if self.secrets is None:
-            self.secrets = self._fetch()
-        return self.secrets
 
-    def refresh(self):
-        """Force refresh of secrets from the vault service."""
-        self.secrets = self._fetch()
+class KeyValueStorage:
```
bybash: Let's move this class into a proper new file.

src/writer/vault.py
```diff
@@ -44,40 +46,75 @@ def _get_agent_ids(self):
...
+            "Authorization": f"Bearer {self.api_key}",
             "X-Organization-Id": org_id,
             "X-Agent-Id": agent_id,
         }
 
+        response = request_func(headers=headers, timeout=3)
+        response.raise_for_status()
+        return response
+
+writer_kv_storage = KeyValueStorage()
```
bybash: Can we make this one with a singleton pattern?

bybash (changes requested):
Nice one 👍 
A few code-rabbit comments are doable and added some comments 🙏

src/writer/vault.py
UladzislauK-Writer: It's probably overcomplicating things. We rely on this pattern throughout the repo without the guardrails. And we don't need them since this class is essentially stateless

PR #1146, opened 2025-10-27
feat: notify AM repo when WF is published

UladzislauK-Writer:
## Summary by CodeRabbit

* **Chores**
  * Publishing workflow now runs from the dev branch, computes an RC version automatically (queries package index and repo state), selects a final version, and can create/push annotated RC tags for dev releases.
  * Packaging and publish steps apply the computed or provided final version for build and release.
  * Cross-workflow trigger accepts dynamic event types and merged payloads (including commit reference), logs inputs, and dispatches using the combined payload.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,70 +2,123 @@ name: publish
 
 on:
     push:
+        branches:
+            - 'dev'
         tags:
-            - 'agent-builder-turbo-[0-9]+.[0-9]+.[0-9]+-image-[0-9]+.[0-9]+.[0-9]+-main'
-            - 'agent-builder-turbo-[0-9]+.[0-9]+.[0-9]+rc[0-9]+-image-[0-9]+.[0-9]+.[0-9]+-main'
+            - 'writer-framework-[0-9]+.[0-9]+.[0-9]+'
     workflow_dispatch:
         inputs:
-            tag:
-                description: 'Tag to publish'
+            commit_sha:
+                description: 'Commit SHA to publish'
                 required: true
             version:
-                description: 'Version to set before publishing (optional, overrides extracted version)'
+                description: 'Version to publish (optional, overrides calculated version)'
                 required: false
 
 jobs:
     build:
         runs-on: ubuntu-latest
         timeout-minutes: 10
+        outputs:
+            version: ${{ steps.set_version.outputs.version }}
 
         steps:
             - uses: actions/checkout@v4
               with:
-                ref: ${{ github.event.inputs.tag || github.ref_name }}
-
-            - name: Extract version from tag
-              id: extract_version
-              run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -2,70 +2,123 @@ name: publish
 
 on:
     push:
+        branches:
+          - 'dev'
         tags:
-            - 'agent-builder-turbo-[0-9]+.[0-9]+.[0-9]+-image-[0-9]+.[0-9]+.[0-9]+-main'
-            - 'agent-builder-turbo-[0-9]+.[0-9]+.[0-9]+rc[0-9]+-image-[0-9]+.[0-9]+.[0-9]+-main'
+         - 'writer-framework-[0-9]+.[0-9]+.[0-9]+'
```
as-flow: Need indentation here

as-flow: Do we not need rc version here?

.github/workflows/trigger-workflow.yml
```diff
@@ -1,9 +1,16 @@
...
 on:
-  push:
-    branches:
-      - 'dev'
+  workflow_call:
+    inputs:
+      event_type:
+        description: 'Event type to send to the agent manager'
+        required: true
+        type: string
```
as-flow: Can we use a choice here? Will be more self documenting

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs

.github/workflows/trigger-workflow.yml
```diff
@@ -17,13 +24,22 @@ jobs:
       run: |
         repo_owner="WriterInternal"
         repo_name="be.agent-manager"
-        event_type="trigger-workflow"
-        commit_sha=${{ github.sha }}
+        event_type="${{ github.event.inputs.event_type || 'framework_updated' }}"
```
as-flow: We should investigate these code injection comments and make sure we don't expose any vulnerabilities

as-flow (changes requested):

.github/workflows/publish.yml
UladzislauK-Writer: No, it's designed to specifically catch final releases

rc's we autogenerate just from repo commits, but for releases we still need human intervention (merging PR in the AM) which then creates a tag, which then triggers publishing

.github/workflows/trigger-workflow.yml
UladzislauK-Writer: For this specific case they are basically useless

You can't inject `github.sha`. `event_type` and `extra_payload` are fully controlled by us

.github/workflows/trigger-workflow.yml
UladzislauK-Writer: I used workflow_call, not workflow_dispatch

image

PR #1149, opened 2025-10-28
feat: Add tab change event to tab container + binding for active tab

sheikhuzairhussain:
## Summary by CodeRabbit

* **New Features**
  * New wf-tab-change event emitted when the active tab changes for reactive integrations.
  * Tabs now initialize to a sensible visible/default tab on mount and stay synchronized with form-driven state.
  * Root container references added to improve DOM interactions and lifecycle handling.

* **Bug Fixes**
  * More reliable tab registration, activation and visibility handling for correct switching behavior.

* **Documentation**
  * Added event documentation and example handler clarifying payload and usage.

### DIFF
--- a/src/ui/src/components/core/layout/CoreTab.vue
+++ b/src/ui/src/components/core/layout/CoreTab.vue
@@ -75,11 +75,10 @@ export default {
 };
 </script>
 <script setup lang="ts">
-import { computed, inject, onBeforeMount, useTemplateRef, watch } from "vue";
+import { computed, inject, onBeforeMount, watch } from "vue";
 import injectionKeys from "@/injectionKeys";
 import BaseContainer from "../base/BaseContainer.vue";
 
-const rootEl = useTemplateRef("rootEl");
 const fields = inject(injectionKeys.evaluatedFields);
 const instancePath = inject(injectionKeys.instancePath);
 const instanceData = inject(injectionKeys.instanceData);
@@ -147,10 +146,8 @@ const getMatchingTabInstancePath = () => {
 
 const activateTab = () => {
 	const tabContainerData = getTabContainerData();
-	tabContainerData.value = {
-		activeTab: getMatchingTabInstancePath(),
-		activeTabName: fields.name.value,
-	};
+	tabContainerData.value.activeTab = getMatchingTabInstancePath();
+	tabContainerData.value.activeTabName = fields.name.value;
 };
 
 const checkIfTabIsParent = (childId: Component["id"]): boolean => {
--- a/src/ui/src/components/core/layout/CoreTab.vue
+++ b/src/ui/src/components/core/layout/CoreTab.vue
@@ -149,18 +149,8 @@ const activateTab = () => {
 	const tabContainerData = getTabContainerData();
 	tabContainerData.value = {
 		activeTab: getMatchingTabInstancePath(),
+		activeTabName: fields.name.value,
 	};
-
-	if (rootEl.value) {
-		const payload = fields.name.value;
-		const event = new CustomEvent("wf-tab-change", {
-			detail: {
-				payload,
-			},
-			bubbles: true,
-		});
-		rootEl.value.dispatchEvent(event);
-	}
 };
 
 const checkIfTabIsParent = (childId: Component["id"]): boolean => {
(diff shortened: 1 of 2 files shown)


src/ui/src/components/core/layout/CoreTabs.vue
```diff
@@ -36,6 +43,14 @@ export default {
 		description,
 		category: "Layout",
 		allowedChildrenTypes: ["tab", "repeater"],
+		events: {
+			"wf-tab-change": {
+				desc: "Sent when the active tab changes.",
+				stub: tabChangeHandlerStub.trim(),
+				eventPayloadExample: "Tab Name",
+				bindable: true,
```
madeindjs, line 52: You didn't implement the bindable mode. Bindable mode means the current tab can be synced to a state variable, which is not the case here.

You can simply remove it from now

```suggestion
```

madeindjs (changes requested):
I have this error when trying to use the event

```
  File "/Users/alexandrerousseau/github/writer/writer-framework@upstream/src/writer/blueprints.py", line 776, in _execute
    raise BlueprintExecutionError(
        f"Blueprint execution was stopped due to an error - {e.__class__.__name__}: {e}"
    ) from e
writer.ss_types.BlueprintExecutionError: Blueprint execution was stopped due to an error - WriterConfigurationError: The field `message` is required. It was left empty.
```

madeindjs (approved):

src/ui/src/components/core/layout/CoreTabs.vue
```diff
@@ -36,6 +43,13 @@ export default {
 		description,
 		category: "Layout",
 		allowedChildrenTypes: ["tab", "repeater"],
+		events: {
+			"wf-tab-change": {
+				desc: "Sent when the active tab changes.",
+				stub: tabChangeHandlerStub.trim(),
+				eventPayloadExample: "Tab Name",
+			},
```
bybash, line 53: Wdyt about adding `bindable: true`? @madeindjs @sheikhuzairhussain

madeindjs: I think it's possible, but you'll need to `watch` the field and update the active tab accordingly. It can be a good exercice to learn the codebase if you feel it @sheikhuzairhussain .

sheikhuzairhussain:
@madeindjs @bybash Please re-review the PR when you get time. The active tab is now bindable to a state variable.

src/ui/src/components/core/layout/CoreTab.vue
```diff
@@ -180,12 +182,13 @@ const isTabActive = computed(() => {
 });
 
 onBeforeMount(() => {
-	if (isTabBit.value) return;
+	if (!isTabBit.value) return;
```
madeindjs, line 182: Why did you need to invert this if statement?

src/ui/src/components/core/layout/CoreTab.vue
```diff
@@ -146,9 +146,8 @@ const getMatchingTabInstancePath = () => {
 
 const activateTab = () => {
 	const tabContainerData = getTabContainerData();
-	tabContainerData.value = {
-		activeTab: getMatchingTabInstancePath(),
-	};
+	tabContainerData.value.activeTab = getMatchingTabInstancePath();
+	tabContainerData.value.activeTabName = fields.name.value;
```
bybash, line 150:
```suggestion
	tabContainerData.value = {
	      activeTab: getMatchingTabInstancePath();
	      activeTabName: fields.name.value;
	 }
```

madeindjs (approved):

src/ui/src/components/core/layout/CoreTab.vue
```diff
@@ -180,12 +179,13 @@ const isTabActive = computed(() => {
...
 	const tabContainerData = getTabContainerData();
-	const activeTab = tabContainerData.value?.activeTab;
-	if (activeTab) return;
-	if (!isComponentVisible(componentId, instancePath)) return;
-	tabContainerData.value = { activeTab: instancePath };
+	tabContainerData.value.tabs.push({
+		name: fields.name.value,
+		instancePath: getMatchingTabInstancePath(),
+		componentId: componentId,
+	});
```
bybash: What is the reason here? 
Also I think use mutations not directly push.


```suggestion
	tabContainerData.value.tabs = [...tabContainerData.value.tabs, {
		name: fields.name.value,
		instancePath: getMatchingTabInstancePath(),
		componentId: componentId,
	}]
```

sheikhuzairhussain: > What is the reason here?

We use the list of tabs heavily in `CoreTabs.vue`

> Also I think use mutations not directly push.

We don't really need to spread the array because Vue can reliably detect array push operations ([reference](https://vuejs.org/guide/essentials/reactivity-fundamentals.html#deep-reactivity))

src/ui/src/components/core/layout/CoreTab.vue
sheikhuzairhussain: This is a completely different handler now. This logic was moved to `CoreTabs.vue`.

src/ui/src/components/core/layout/CoreTab.vue
sheikhuzairhussain: If we reassign `tabContainerData.value` to a new object, it severs the connection to the original state object that the parent `CoreTabs.vue` is still watching. The parent and child components are reference two different objects, which breaks the two-way binding

PR #1156, opened 2025-10-31
fix: update reasoning tool descriptions and finalization protocol in …

as-flow:
…WriterToolCalling

Refs: AB-515

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified reasoning prompts to emphasize brief summaries and explicit completion signaling.
  * Introduced a finalization protocol to drive consistent “DONE” behavior and finish outputs reliably.
  * Improved run-time flow to prefer final message content, allow a last iteration when needed, and fallback to last available content.
  * Enhanced state tracking to more robustly mark completion and capture results.

### DIFF
--- a/src/writer/blocks/writertoolcalling.py
+++ b/src/writer/blocks/writertoolcalling.py
@@ -147,20 +147,20 @@ class WriterToolCalling(WriterBlock):
         reasoning_tool = {
             "type": "function",
             "name": "disclose_reasoning",
-            "description": "Use this function to disclose your reasoning.",
+            "description": "Use this only to briefly summarize your reasoning and to signal completion. Call this when you're ready to finalize your answer.",
             "callable": reasoning_callable,
             "parameters": {
                 "thought": {
                     "type": "string",
-                    "description": "A look into your internal reasoning.",
+                    "description": "A brief summary of your reasoning process.",
                 },
                 "action": {
                     "type": "string",
-                    "description": "A summary of the actions you took and why.",
+                    "description": "When status is DONE, always set this to 'finalize_and_answer'. Otherwise, summarize the actions you took and why.",
                 },
                 "status": {
                     "type": "string",
-                    "description": "Set to DONE if you consider the task complete. Set to INCOMPLETE if you wish to keep iterating.",
+                    "description": "Set to DONE if you consider the task complete and are ready to provide the final answer. Set to INCOMPLETE if you wish to keep iterating.",
                 },
             },
         }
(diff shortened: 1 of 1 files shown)


src/writer/blocks/writertoolcalling.py
```diff
@@ -172,9 +172,33 @@ def reasoning_callable(**kwargs):
...
+            2. **IMMEDIATELY AFTER** that tool call, output a normal assistant message (no further tool calls) containing your complete, final answer to the user.
+            
+            3. **DO NOT** end with a null/None action. Always provide the full answer in your final assistant message.
+            
+            If more information is still needed, call `disclose_reasoning` with `"status": "INCOMPLETE"` and continue gathering information.
+            
+            ## Important Notes
+            - Use tools to gather all necessary information before finalizing
+            - The `action` field in `disclose_reasoning` should always be set to "finalize_and_answer" when status is "DONE"
+            - Your final assistant message (after the final `disclose_reasoning` call) should contain the complete response to the user
```
madeindjs, line 203: Since you use Markdown, I think it might be better to use codeblock for the example

```suggestion
            ## Available Tools

            You have access to various tools to help complete the task. Use them as needed to gather information.

            - `disclose_reasoning` - Use this only to briefly summarize your reasoning and to signal completion

            ## Finalization Protocol (CRITICAL)

            When your analysis is complete, follow these steps EXACTLY:

            1. Make one final function call to `disclose_reasoning` with:
            ```json
            {
              "thought": "<brief reasoning summary>",
              "action": "finalize_and_answer",
              "status": "DONE"
            }
            ```
            2. **IMMEDIATELY AFTER** that tool call, output a normal assistant message (no further tool calls) containing your complete, final answer to the user.
            3. **DO NOT** end with a null/None action. Always provide the full answer in your final assistant message.

            If more information is still needed, call `disclose_reasoning` with `"status": "INCOMPLETE"` and continue gathering information.

            ## Important Notes

            - Use tools to gather all necessary information before finalizing
            - The `action` field in `disclose_reasoning` should always be set to "finalize_and_answer" when status is "DONE"
            - Your final assistant message (after the final `disclose_reasoning` call) should contain the complete response to the user

```

madeindjs (approved):

src/writer/blocks/writertoolcalling.py
as-flow: Thanks. Updated

src/writer/blocks/writertoolcalling.py
```diff
@@ -203,9 +230,21 @@ def run(self):
...
+                    # If no content yet, allow one more iteration to get the final message
+                    # (the tool call processing might need another round)
+                    if i < max_iterations - 1:
+                        continue
                     break
 
-            self.result = msg.get("content")
+            # If we didn't capture the result yet, try to get it from the last message
+            if not self.result:
+                self.result = msg.get("content", "")
```
bybash, line 247: What about returning error if there is no answer in the end?

src/writer/blocks/writertoolcalling.py
```diff
@@ -203,9 +230,21 @@ def run(self):
...
                 if self.is_complete:
+                    # According to the protocol, after disclose_reasoning with status="DONE",
+                    # the agent should output a final assistant message with the complete answer.
+                    # Check if this message has content (the final answer)
+                    if msg.get("content"):
+                        self.result = msg.get("content")
+                        break
+                    # If no content yet, allow one more iteration to get the final message
+                    # (the tool call processing might need another round)
+                    if i < max_iterations - 1:
```
bybash, line 241: Are we sure that when status is DONE, will next iteration change anything?

as-flow: Sometimes the status would be set to Done, but the output would be null. We're specifically guarding against that edge case.

src/writer/blocks/writertoolcalling.py
as-flow: That might be a breaking change because we didn't raise an error before

PR #1157, opened 2025-11-03
feat(ui): close socket on inactivity - AB-591

madeindjs:
https://github.com/user-attachments/assets/4b61374f-80f7-4c84-b580-3d99abd3be0a

## Summary by CodeRabbit

* **New Features**
  * Automatic session expiration after a configurable inactivity period.
  * Session-expired modal with a reconnect action.

* **Improvements**
  * More robust WebSocket management to avoid duplicate connections and improve stability.
  * Operations that run or edit content temporarily pause the session timeout to prevent unexpected disconnects.

### DIFF
--- a/src/ui/src/builder/useSocketTimeout.ts
+++ b/src/ui/src/builder/useSocketTimeout.ts
@@ -1,0 +1,71 @@
+import { useAbortController } from "@/composables/useAbortController";
+import type { Core } from "@/writerTypes";
+import { ref, onMounted, watch } from "vue";
+
+/**
+ * @param timeoutMin the inactivity time required to close the socket
+ */
+export function useSocketTimeout(wf: Core, timeoutMin: number) {
+	const timeoutMs = timeoutMin * 60 * 1_000;
+
+	let timer = undefined;
+
+	const socketClosed = ref(false);
+	const reconnecting = ref(false);
+	const prevent = ref(false);
+
+	function schedule() {
+		if (document.visibilityState === "visible") return;
+		clearSchedule();
+		timer = setTimeout(() => {
+			wf.stopSync();
+			socketClosed.value = true;
+		}, timeoutMs);
+	}
+
+	function clearSchedule() {
+		if (timer) clearTimeout(timer);
+	}
+
+	watch(prevent, () => {
+		if (prevent.value) clearSchedule();
+	});
+
+	async function reconnect() {
+		reconnecting.value = true;
+		try {
+			await wf.init();
+			socketClosed.value = false;
+		} catch {
+			window.location.reload(); // fallback to full reload
+		} finally {
+			reconnecting.value = false;
+		}
+	}
+
+	const abort = useAbortController();
+
+	onMounted(() => {
+		document.addEventListener(
+			"visibilitychange",
+			() => {
+				if (document.visibilityState === "visible") {
+					clearSchedule();
+				} else if (!prevent.value) {
+					schedule();
+				}
+			},
+			{ signal: abort.signal },
+		);
+	});
+
+	return {
+		timeoutMin,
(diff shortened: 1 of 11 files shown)


src/ui/src/builder/useSocketTimeout.ts
```diff
@@ -0,0 +1,71 @@
...
+			socketClosed.value = true;
+		}, timeoutMs);
+	}
+
+	function clearSchedule() {
+		if (timer) clearTimeout(timer);
+	}
+
+	watch(prevent, () => {
+		if (prevent.value) clearSchedule();
```
UladzislauK-Writer, line 31: Shouldn't we have an `else` here? So when we finish running a blueprint (or editing a file) and set `prevent` back to `false` we should schedule the timer again

UladzislauK-Writer:
I can see preventions for file editing and blueprints. What about other activity like editing interface/blueprints? We talked about `useWriterTracking` during the meeting and I don't see `schedule` being called as a result of such actions

madeindjs:
> I can see preventions for file editing and blueprints. What about other activity like editing interface/blueprints? We talked about `useWriterTracking` during the meeting and I don't see `schedule` being called as a result of such actions

I finally went with the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) because I feel it's a more robust and simpler solution.

So, if the user puts the tab in the background (or minimizes the editor), we start a 10-minute counter to close the socket. Then, I prevent the counter from starting if the code panel has unsaved changes.

src/ui/src/builder/useSocketTimeout.ts
madeindjs: No, because we run the `schedule` method when the visibility changes (like when th browser tab is put in the background).

UladzislauK-Writer (approved):

as-flow:
Does this affect the collaboration feature in any way? So for example if a user disconnects from the app while another user is working on it in another tab. (Or it could be the same user who has two tabs open by accident). If they go back to the old tab and perform an action, would they basically roll back to the inactive user's state? Is it better to force a refresh if so?

madeindjs:
> Does this affect the collaboration feature in any way? So for example if a user disconnects from the app while another user is working on it in another tab. (Or it could be the same user who has two tabs open by accident). If they go back to the old tab and perform an action, would they basically roll back to the inactive user's state? Is it better to force a refresh if so?

@as-flow , good question. The reconnect feature will try to call `wf.init()`. This will almost do a full refresh:

1. Call `/init` to start a new session and get a fresh list of components.
2. Establish a new socket connection.

So it means that every change done during his offline time will be restored. If it's not possible, we fallback to a full reload

https://github.com/writer/writer-framework/blob/3cf4e08290d6e3c0200edbd618fd7d1e19707876/src/ui/src/builder/useSocketTimeout.ts#L34-L40

as-flow:
> > Does this affect the collaboration feature in any way? So for example if a user disconnects from the app while another user is working on it in another tab. (Or it could be the same user who has two tabs open by accident). If they go back to the old tab and perform an action, would they basically roll back to the inactive user's state? Is it better to force a refresh if so?
> 
> @as-flow , good question. The reconnect feature will try to call `wf.init()`. This will almost do a full refresh:
> 
> 1. Call `/init` to start a new session and get a fresh list of components.
> 2. Establish a new socket connection.
> 
> So it means that every change done during his offline time will be restored. If it's not possible, we fallback to a full reload
> 
> https://github.com/writer/writer-framework/blob/3cf4e08290d6e3c0200edbd618fd7d1e19707876/src/ui/src/builder/useSocketTimeout.ts#L34-L40

Thanks!

PR #1161, opened 2025-11-04
[AB-72] Hold mouse to select multiple blocks

rodionchernyshov:
## Summary by CodeRabbit

* **New Features**
  * Drag-to-select with a visible selection rectangle and cursor feedback (crosshair when selecting; grab/grabbing when panning)
  * Selection overlay and unified click/drag controls across builder and blueprint views

* **Bug Fixes / Improvements**
  * Prevents accidental single-click selection during multi-selection and after incomplete drags
  * More robust hit-testing that respects zoom/scroll and cleans up document-level event handling to avoid errant interactions

### DIFF
--- a/src/ui/src/builder/BuilderApp.vue
+++ b/src/ui/src/builder/BuilderApp.vue
@@ -365,8 +365,6 @@ function handleRendererClick(ev: PointerEvent): void {
 	);
 	if (unselectableEl) return;
 
-	// In blueprints mode, if we have multiple selections, don't allow single-click selection
-	// This prevents interference with drag-to-select multi-selection
 	if (
 		builderMode.value === "blueprints" &&
 		ssbm.selectionStatus.value === SelectionStatus.Multiple
@@ -375,7 +373,6 @@ function handleRendererClick(ev: PointerEvent): void {
 			"[data-writer-id]",
 		);
 		if (targetEl && ssbm.isComponentIdSelected(targetEl.dataset.writerId)) {
-			// If clicking on an already-selected component in multi-selection mode, don't change selection
 			ev.preventDefault();
 			ev.stopPropagation();
 			return;
--- a/src/ui/src/builder/BuilderApp.vue
+++ b/src/ui/src/builder/BuilderApp.vue
@@ -24,15 +24,25 @@
 						<BuilderVault v-if="builderMode === 'vault'" />
 						<div
 							v-else
+							ref="rendererWrapperEl"
 							class="rendererWrapper"
 							:class="{
 								addNoteCursor:
 									notesManager.isAnnotating.value &&
 									ssbm.mode.value !== 'preview',
+								isSelecting: isCursorSelecting,
+								canSelect:
+									isHoveringSelectableArea &&
+									!isCursorSelecting,
 							}"
 							@scroll="refreshNotesPosition"
+							@mousedown.capture="handleRendererMousedown"
+							@mousemove="handleRendererMousemove"
+							@mouseup="handleRendererMouseup"
+							@mouseleave="handleRendererMouseleave"
 						>
 							<ComponentRenderer
+								ref="rendererEl"
 								class="componentRenderer"
 								:class="{
 									settingsOpen: ssbm.isSingleSelectionActive,
@@ -45,6 +55,17 @@
 								@dblclick="handleRendererDblClick"
 							>
 							</ComponentRenderer>
+							<!-- Selection rectangle overlay -->
+							<div
+								v-if="selectionRect.isSelecting"
+								class="selectionRectangle"
+								:style="{
+									left: `${selectionRect.left}px`,
+									top: `${selectionRect.top}px`,
+									width: `${selectionRect.width}px`,
+									height: `${selectionRect.height}px`,
+								}"
+							></div>
 						</div>
 						<BuilderSettings
 							v-if="ssbm.isSingleSelectionActive"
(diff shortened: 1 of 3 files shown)


src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+		isSelecting: false,
+		startX: 0,
+		startY: 0,
+		left: 0,
+		top: 0,
+		width: 0,
+		height: 0,
+	});
+
+	const dragAbortController = new AbortController();
```
madeindjs:
```suggestion
	const dragAbortController = useAbortController();
```

src/ui/src/builder/BuilderApp.vue
madeindjs, line 1: BuilderApp is the top level component for UI, Blueprint & Preview.

Since the drag & drop is only for Blueprint, I think you should call `useDragToSelect` in `BlueprintsBlueprint` or `BlueprintsRoot`

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+	wrapperRef: Ref<HTMLElement | null>;
+	builderMode: Ref<BuilderManagerMode>;
+	builderManager: BuilderManager;
+	isAnnotating: Ref<boolean>;
+}
+
+export function useDragToSelect(options: UseDragToSelectOptions) {
+	const { wrapperRef, builderMode, builderManager, isAnnotating } = options;
+
+	const selectionRect = ref<SelectionRectangle>({
```
madeindjs: You can use `shallowRef` to improve performances

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+
+	function handleMouseleave(_ev: MouseEvent) {
+		isHoveringSelectableArea.value = false;
+	}
+
+	return {
+		selectionRect,
+		isCursorSelecting,
+		isHoveringSelectableArea,
+		isSelecting: computed(() => selectionRect.value.isSelecting),
```
madeindjs:
```suggestion
		isSelecting: isCursorSelecting,
```

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+
+		const nodeContainer =
+			blueprintsBlueprint.querySelector<HTMLElement>(".nodeContainer");
+		return nodeContainer ? nodeContainer.contains(target) : true;
+	}
+
+	function cleanupDragSelection() {
+		if (!selectionRect.value.isSelecting) return;
+
+		selectionRect.value.isSelecting = false;
```
madeindjs: If you use `shallowRef`, you'll need a full reassignment


```suggestion
		selectionRect.value = {
		  isSelecting: false,
	  	startX: 0,
	  	startY: 0,
	  	left: 0,
	  	top: 0,
	  	width: 0,
	  	height: 0,
	 };
```

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+	height: number;
+}
+
+const MIN_SELECTION_SIZE_PX = 5;
+const DRAG_SELECTION_CLICK_DELAY_MS = 100;
+
+interface UseDragToSelectOptions {
+	wrapperRef: Ref<HTMLElement | null>;
+	builderMode: Ref<BuilderManagerMode>;
+	builderManager: BuilderManager;
```
madeindjs: You don't need `builderMode` , you can use `builderManager.mode`

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
+import { ref, computed, type Ref, onUnmounted } from "vue";
+import type { BuilderManagerMode } from "../builderManager";
+import { generateBuilderManager } from "../builderManager";
+
+type BuilderManager = ReturnType<typeof generateBuilderManager>;
```
madeindjs: You can use this type.
```suggestion
import type { BuilderManager } from "@/writerTypes";
```

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+			left: startX,
+			top: startY - wrapper.scrollTop,
+			width: 0,
+			height: 0,
+		};
+
+		document.body.style.userSelect = "none";
+		document.body.style.cursor = "crosshair";
+
+		document.addEventListener("mousemove", handleDocumentMousemove, {
```
madeindjs: 1. You don't remove this event listener on click.
2. Maybe you can listen only to `wrapperRef.value` instead of `document`

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+			isHoveringSelectableArea.value = false;
+			return;
+		}
+
+		if (builderMode.value === "blueprints" && !ev.altKey) {
+			isHoveringSelectableArea.value = false;
+			return;
+		}
+
+		const target = ev.target as HTMLElement;
```
madeindjs: It's better to cast at runtime.

```suggestion
    if (!(ev.target instanceof HTMLElement)) return
		const target = ev.target;
```

src/ui/src/builder/composables/useDragToSelect.ts
```diff
@@ -0,0 +1,341 @@
...
+				builderManager.appendSelection(
+					componentId,
+					instancePath,
+					"click",
+				);
+			});
+			justCompletedDragSelection.value = true;
+			setTimeout(() => {
+				justCompletedDragSelection.value = false;
+			}, DRAG_SELECTION_CLICK_DELAY_MS);
```
madeindjs: I don't get why do we need to wait ? Maybe you want `nextTick`

```suggestion
      nextTick().then(() => justCompletedDragSelection.value = false);
```

It should be a bit better, but it's always a bad idea to wait an arbitrary amount of time.

madeindjs (approved):

PR #1162, opened 2025-11-05
chore: update dependencies

UladzislauK-Writer:
## Chores
* Updated version constraints for core dependencies (authlib, requests, websockets, and numpy) to support a broader range of releases.
* Added explicit version specifications for additional supporting libraries (jinja2, pandas, plotly, pyarrow, and pydantic).


poetry.lock
```diff
@@ -2067,84 +2081,81 @@ files = [
 
 [[package]]
 name = "websockets"
-version = "12.0"
+version = "15.0.1"
```
madeindjs, line 2084: Sounds like a big bump!

From the [changelog](https://websockets.readthedocs.io/en/stable/project/changelog.html#id3) , I see some breaking changes

- Client connections use SOCKS and HTTP proxies automatically.
- Keepalive is enabled in the [threading](https://docs.python.org/3/library/threading.html#module-threading) implementation.
- The new [asyncio](https://docs.python.org/3/library/asyncio.html#module-asyncio) implementation is now the default.

UladzislauK-Writer: I checked everything, we are good to go

PR #1163, opened 2025-11-05
feat: icon loading enhancements

bybash:
Before:
https://www.loom.com/share/f16291c370984500a21ff83bc0e23057

After:
https://www.loom.com/share/1ca8f360798440ba94ae4e40a19d357b

## Summary by CodeRabbit

* **New Features**
  * Added skeleton loading placeholders and configurable loader dimensions for icon/image placeholders to improve visual feedback during loading.
  * Improved image loading logic to skip invalid content and reduce unnecessary re-processing.

* **Style**
  * Adjusted select/trigger label layout to better handle overflow and alignment.

* **Bug Fixes**
  * Corrected a typographical comment.

* **Tests**
  * Updated tests to assert skeleton rendering and final image loading behavior.

### DIFF
--- a/src/ui/src/components/blueprints/abstract/BlueprintsNode.vue
+++ b/src/ui/src/components/blueprints/abstract/BlueprintsNode.vue
@@ -21,7 +21,12 @@
 		</div>
 		<div class="BlueprintsNode__main">
 			<div class="BlueprintsNode__main__title">
-				<SharedImgWithFallback :urls="possibleImageUrls" class="icon" />
+				<SharedImgWithFallback
+					:urls="possibleImageUrls"
+					class="icon"
+					:loader-max-width-px="24"
+					:loader-max-height-px="24"
+				/>
 				<BlueprintsNodeNamer
 					:component-id="componentId"
 					class="nodeNamer"
@@ -525,13 +530,6 @@ watch(isEngaged, () => {
 .BlueprintsNode__main__title img {
 	width: 24px;
 	height: 24px;
-}
-
-.BlueprintsNode__main__title .skeleton {
-	max-width: 24px;
-	max-height: 24px;
-	aspect-ratio: 1 / 1;
-	border-radius: 4px;
 }
 
 .BlueprintsNode__main__title .deprecationNotice {
--- a/src/ui/src/components/blueprints/abstract/BlueprintsNode.vue
+++ b/src/ui/src/components/blueprints/abstract/BlueprintsNode.vue
@@ -527,6 +527,13 @@ watch(isEngaged, () => {
 	height: 24px;
 }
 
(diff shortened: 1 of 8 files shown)


src/ui/src/components/blueprints/abstract/BlueprintsNode.vue
```diff
@@ -527,6 +527,13 @@ watch(isEngaged, () => {
 	height: 24px;
 }
 
+.BlueprintsNode__main__title .skeleton {
+	max-width: 24px;
+	max-height: 24px;
+	aspect-ratio: 1 / 1;
+	border-radius: 4px;
+}
```
madeindjs: This should not work. The CSS here is scoped to the component, so `.skeleton` is not accessible.

You could use `:deep` to traverse the scope.

```suggestion
.BlueprintsNode__main__title :deep(.skeleton) {
	max-width: 24px;
	max-height: 24px;
	aspect-ratio: 1 / 1;
	border-radius: 4px;
}
```

But I think the better solution could be to add a prop for `SharedImgWithFallback`. So maybe something like

```vue
<SharedImgWithFallback.vue :url loader-max-width-px="24" loader-max-height-px="24" />
```

bybash: Yes I was thinking about this too, let me check this.
```<SharedImgWithFallback.vue :url loader-max-width-px="24" loader-max-height-px="24" />```

File: src/ui/src/components/shared/SharedImgWithFallback.vue [hunk 2]
```diff
@@ -1,31 +1,59 @@
...
-			return (src.value = url);
+
+			src.value = url;
+			break;
 		}
+
+		isLoading.value = false;
 	},
 	{ immediate: true },
 );
```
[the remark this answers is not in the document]
bybash: Makes sense 👍

src/ui/src/components/shared/SharedImgWithFallback.vue
```diff
@@ -1,31 +1,87 @@
...
+	return a.every((url, index) => url === b[index]);
+}
+
 watch(
 	toRef(props, "urls"),
 	async (urls) => {
+		// Skip if URLs haven't changed
+		if (areUrlsEqual(urls, previousUrls.value)) {
+			return;
+		}
```
madeindjs, line 56: I'm not sure if you need to use a special ref for `previousUrls`, you can use the previous value from the watcher callback

```suggestion
	async (urls, previousUrls) => {
		// Skip if URLs haven't changed
		if (areUrlsEqual(urls, previousUrls)) {
			return;
		}
```

madeindjs (approved):

PR #1166, opened 2025-11-06
feat(AB-613): Add endpoints for fetching and executing cron branches

sheikhuzairhussain:
## Summary by CodeRabbit

* **New Features**
  * Blueprint execution flow now supports branching to enable conditional workflows
  * New API endpoint available to retrieve all configured cron triggers
  * Enhanced status notifications now distinguish between blueprint and branch execution states


uv.lock
as-flow, line 1: Let's delete this since we use poetry in this repo

src/writer/serve.py
```diff
@@ -379,8 +379,37 @@ async def get_blueprints(request: Request):
 
         return JSONResponse(content=blueprints)
 
+    @app.get("/private/api/blueprint/{blueprint_id}/cron-triggers")
```
as-flow: Let's update test_serve as well with this additional logic

as-flow (changes requested):

src/writer/serve.py
```diff
@@ -379,8 +379,30 @@ async def get_blueprints(request: Request):
...
+        """
+        if not app_runner.bmc_components:
+            return JSONResponse(content=[], status_code=200)
+
+        cron_triggers = [
+            {
+                "id": comp.get("id"),
+                "blueprint_id": comp.get("parentId"),
+                "name": comp.get("name") or comp.get("content", {}).get("name"),
+                "cron_expression": comp.get("content", {}).get("cronExpression"),
```
bybash:
```suggestion
                "cron_expression": comp.get("content", {}).get("cronExpression", ""),
```

uv.lock
sheikhuzairhussain: Oops, pushed it accidentally. I use poetry as a uv tool. Removed in 2e64169722011696733fd4f99e8a5c9d997672c4

src/writer/serve.py
sheikhuzairhussain: Added a test case here: 6cd5cbb612043360432a9cb81a66c8ae7c336f22

src/writer/serve.py
```diff
@@ -379,8 +379,30 @@ async def get_blueprints(request: Request):
...
+            }
+            for comp in app_runner.bmc_components.values()
+            if comp.get("type") == "blueprints_crontrigger"
+        ]
+
+        return JSONResponse(content=cron_triggers, status_code=200)
+
     @app.post("/private/api/blueprint/{blueprint_id}")
-    async def create_blueprint_job(blueprint_id: str, request: Request, response: Response):
+    async def create_blueprint_job(blueprint_id: str, request: Request, response: Response, branch_id: str | None = None):
```
bybash, line 405:
```suggestion
    async def create_blueprint_job(blueprint_id: str, request: Request, response: Response, cron_trigger_id: str | None = None):
```

src/writer/serve.py
```diff
@@ -461,29 +483,50 @@ async def event_logic(queue: asyncio.Queue):
...
+                        }))
+                        return
+
+                await queue.put(await format_event("status", {"status": "executing", "msg": (f"Executing branch: {branch_id}..." if branch_id else f"Executing blueprint: {blueprint_id}...")}))
+
+                if branch_id:
+                    task = asyncio.create_task(
+                        app_runner.handle_event(
+                            session_id,
+                            WriterEvent(
```
bybash, line 508: Who will get this event?

sheikhuzairhussain: This event is handled by the `run_blueprint_branch` method in `core.py`:
https://github.com/writer/writer-framework/blob/ed45bec4e0b63c994ff60a4402deeca5f16187a6/src/writer/core.py#L1293-L1312

src/writer/serve.py
sheikhuzairhussain: The convention across the codebase seems to be `branch_id` and also, this endpoint works for any branch starting with any arbitrary node

src/writer/serve.py
sheikhuzairhussain: Resolved in 0d89bb6ac1a39456b4e857b625e756ef1503d13b

bybash (approved):

PR #1179, opened 2025-11-12
fix(ui): Prevent the socket from being closed during activity - AB-680

madeindjs:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent socket closing while frontend messages are pending to avoid interrupted updates.
  * Shortened default socket timeout to 10s for faster reconnection and improved responsiveness.
  * Improved connection error/close logging for clearer diagnostics.

* **Chores**
  * Removed a runtime timeout toggle to simplify timeout behavior and make connections more predictable.

* **Tests**
  * Added comprehensive tests for scheduling, visibility handling, reconnection flow, and timer behavior.

### DIFF
--- a/src/ui/src/builder/useSocketTimeout.ts
+++ b/src/ui/src/builder/useSocketTimeout.ts
@@ -1,7 +1,7 @@
 import { useAbortController } from "@/composables/useAbortController";
 import { useLogger } from "@/composables/useLogger";
 import type { Core } from "@/writerTypes";
-import { ref, onMounted, watch } from "vue";
+import { ref, onMounted, watch, computed } from "vue";
 
 /**
  * @param timeoutMin the inactivity time required to close the socket
@@ -16,23 +16,55 @@ export function useSocketTimeout(wf: Core, timeoutMin: number) {
 	const reconnecting = ref(false);
 	const prevent = ref(false);
 
+	const ignoreMessageType = new Set([
+		"collaborationPing",
+		"listResources",
+		"keepAlive",
+	]);
+
+	const isMessagePending = computed(() => {
+		for (const a of wf.frontendMessageMap.value.values()) {
+			if (!ignoreMessageType.has(a.type)) return true;
+		}
+		return false;
+	});
+
+	const canCloseSocket = computed(
+		() => !prevent.value && !isMessagePending.value,
+	);
+
+	const abort = useAbortController();
+
 	function schedule() {
-		if (document.visibilityState === "visible") return;
+		if (!canCloseSocket.value || document.visibilityState !== "hidden")
+			return false;
 		clearSchedule();
+		logger.log(`[SocketTimeout] planning timeout in ${timeoutMin}min`);
 		timer = setTimeout(() => {
-			logger.warn(`[SocketTimeout] Closing socket after ${timeoutMin} minutes of inactivity (tab hidden)`);
+			if (!canCloseSocket.value) return;
+			logger.warn(
+				`[SocketTimeout] Closing socket after ${timeoutMin} minutes of inactivity (tab hidden)`,
+			);
 			wf.stopSync();
 			socketClosed.value = true;
 			logger.info(`[SocketTimeout] Socket closed`);
 		}, timeoutMs);
+		return true;
 	}
 
 	function clearSchedule() {
-		if (timer) clearTimeout(timer);
+		if (!timer) return;
+		logger.log(`[SocketTimeout] canceling timeout `);
+		clearTimeout(timer);
+		timer = undefined;
 	}
 
-	watch(prevent, () => {
-		if (prevent.value) clearSchedule();
+	watch(canCloseSocket, () => {
+		if (!canCloseSocket.value) {
+			clearSchedule();
+		} else if (document.visibilityState === "hidden") {
+			schedule(); // Reschedule now that activity is complete
+		}
 	});
 
 	async function reconnect() {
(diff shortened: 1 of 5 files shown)


src/ui/src/builder/useSocketTimeout.ts
```diff
@@ -16,23 +16,36 @@ export function useSocketTimeout(wf: Core, timeoutMin: number) {
...
+	const cantCloseSocket = computed(
+		() => prevent.value || isMessagePending.value,
+	);
+
 	function schedule() {
 		if (document.visibilityState === "visible") return;
 		clearSchedule();
 		timer = setTimeout(() => {
-			logger.warn(`[SocketTimeout] Closing socket after ${timeoutMin} minutes of inactivity (tab hidden)`);
+			if (cantCloseSocket.value) return;
```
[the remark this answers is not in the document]
madeindjs: I don't want to reschedule automatically.

as-flow (changes requested):
We'll have to filter keepAlive from the message map
Worth checking what other messages we'd need to filter

src/ui/src/builder/useSocketTimeout.ts
```diff
@@ -43,34 +66,36 @@ export function useSocketTimeout(wf: Core, timeoutMin: number) {
...
-			() => {
-				if (document.visibilityState === "visible") {
-					clearSchedule();
-				} else if (!prevent.value) {
-					schedule();
-				}
-			},
-			{ signal: abort.signal },
-		);
+		document.addEventListener("visibilitychange", onVisibilityChange, {
```
bybash, line 97: don't we need to remove listener when component is unmounted?

src/ui/src/builder/useSocketTimeout.ts
```diff
@@ -16,23 +16,46 @@ export function useSocketTimeout(wf: Core, timeoutMin: number) {
...
+	const canCloseSocket = computed(
+		() => !prevent.value && !isMessagePending.value,
+	);
+
+	const abort = useAbortController();
+
 	function schedule() {
-		if (document.visibilityState === "visible") return;
+		if (!canCloseSocket.value || document.visibilityState !== "hidden")
+			return false;
```
bybash, line 40: Do you need a return value?

src/ui/src/builder/useSocketTimeout.ts
```diff
@@ -16,23 +16,46 @@ export function useSocketTimeout(wf: Core, timeoutMin: number) {
 	const reconnecting = ref(false);
 	const prevent = ref(false);
 
+	const isMessagePending = computed(
+		() => wf.frontendMessageMap.value.size > 0,
```
bybash: Will this let us know a BP running?

src/ui/src/builder/useSocketTimeout.ts
madeindjs: We use a signal, which means the listener will be closed when the `AbortController` is aborted 

https://github.com/writer/writer-framework/blob/655fc69622f39914e6c22d417f9b56eab6d10b97/src/ui/src/builder/useSocketTimeout.ts#L89-L89

src/ui/src/builder/useSocketTimeout.ts
madeindjs: Yes, I'm using it for unit test 

https://github.com/writer/writer-framework/blob/655fc69622f39914e6c22d417f9b56eab6d10b97/src/ui/src/builder/useSocketTimeout.test.ts#L73-L73

src/ui/src/builder/useSocketTimeout.ts
madeindjs: Yes, running a blueprint will send a socket message `event`
```json
{
    "type": "event",
    "trackingId": 6,
    "payload": {
        "type": "wf-run-blueprint",
    }
}
```
And then we wait for the response in the socket, which happens after the end of the blueprint execution.
```json
{
  "messageType": "eventResponse",
  "trackingId": 6,
  "payload": {
    "result": { "ok": true, "result": null }
  }
}
```

Also, meanwhile we use the `setupMessageFollowUp` func to send a `stateEnquiry` to check the state of the session, and the log of the blueprint every 150 ms.


So, during a blueprint execution, we'll have the `stateEnquiry|event` message in `frontendMessageMap`

madeindjs:
> We'll have to filter keepAlive from the message map Worth checking what other messages we'd need to filter

@as-flow , you made a good point. i fixed it like this
https://github.com/writer/writer-framework/blob/be3967f977636d5c87c0cd96342d07491dcc1bbd/src/ui/src/builder/useSocketTimeout.ts#L19-L30

src/ui/src/builder/useSocketTimeout.ts
```diff
@@ -16,23 +16,55 @@ export function useSocketTimeout(wf: Core, timeoutMin: number) {
...
 
+	const ignoreMessageType = new Set([
+		"collaborationPing",
+		"listResources",
+		"keepAlive",
+	]);
+
+	const isMessagePending = computed(() => {
+		for (const a of wf.frontendMessageMap.value.values()) {
+			if (!ignoreMessageType.has(a.type)) return true;
```
as-flow, line 27: Would it be easy to add a test for this logic? Making sure that keepAlive for example doesn't reset the timer

as-flow (approved):

src/ui/src/builder/useSocketTimeout.ts
madeindjs: Implemented in https://github.com/writer/writer-framework/pull/1179/commits/b4a29c730540122eed6e92715e1da97aca00e976

PR #1181, opened 2025-11-13
fix: update loading animation

rodionchernyshov:
Screenshot 2025-11-13 at 12 18 18

## Summary by CodeRabbit

* **Refactor**
  * Replaced the SVG loader with a CSS-based dot animation (same appearance, dynamic color/size) for smoother visuals and easier maintenance.
  * Loader component is now included directly (no lazy wrapper), simplifying behavior and loading timing; public component props unchanged.

### DIFF
--- a/src/ui/src/wds/WdsButton.vue
+++ b/src/ui/src/wds/WdsButton.vue
@@ -28,11 +28,7 @@ export type WdsButtonSize = "big" | "small" | "icon" | "smallIcon";
 
 <script setup lang="ts">
 import { computed, CSSProperties, PropType } from "vue";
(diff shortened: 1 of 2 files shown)


madeindjs (commented):
I'm not sure that it's solving the issue. The weird effect is because we use Async Component. 

https://github.com/writer/writer-framework/blob/d5940bdeb9d8d576d70fb3c94c1c0b525f64e623/src/ui/src/wds/WdsButton.vue#L33-L35

It means Vue.js will split the bundle and create a JS chunk for the `WdsLoaderDots` . So, to display the loader icons, the browser will do a fetch request and display the skeleton loader meanwhile.

bybash (approved):

PR #1182, opened 2025-11-13
fix(api): enhance import endpoint logs - AB-561

as-flow:
## Release Notes

* **Bug Fixes**
  * Enhanced ZIP import functionality with improved error handling for more reliable file uploads and processing
  * Temporary files are now properly managed during import operations
  * Import processes provide explicit success confirmation upon completion
  * Invalid import requests return clearer, more informative error messages

### DIFF
--- a/src/writer/serve.py
+++ b/src/writer/serve.py
@@ -262,6 +262,7 @@ def get_asgi_app(
             raise HTTPException(status_code=400, detail="Only .zip files are supported.")
 
         MAX_FILE_SIZE = 200 * 1024 * 1024
+        tmp_path = None
 
         try:
             with tempfile.NamedTemporaryFile(delete=False) as tmp:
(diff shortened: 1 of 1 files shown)


UladzislauK-Writer:
We can use that but I doubt this will solve the root cause

We get 502 when proxy couldn't make a request to the instance. In most cases this means that nothing was serving on port 8080. Import restarts `writer` process, so we have some downtime. Most likely we get a request during that downtime

src/writer/serve.py
```diff
@@ -275,10 +276,21 @@ async def import_zip(file: UploadFile = File(...)):
...
+            except Exception:
+                logging.warning("Import succeeded but post-import setup encountered an issue", exc_info=True)
+        finally:
+            if tmp_path and os.path.exists(tmp_path):
+                try:
+                    os.remove(tmp_path)
+                except Exception:
+                    logging.warning("Failed to remove temp file, but import succeeded")
+        
+        return {"success": True, "message": "Import completed successfully"}
```
bybash, line 293: Is this correct? Shouldn't it be after os.remove call?

as-flow:
> We can use that but I doubt this will solve the root cause
> 
> We get 502 when proxy couldn't make a request to the instance. In most cases this means that nothing was serving on port 8080. Import restarts `writer` process, so we have some downtime. Most likely we get a request during that downtime

Do you have a proposed solution? The error is Failed to import the project: Failed to connect to import API

Sometimes I get 500 and sometimes 502

Also I can't reproduce when running framework locally. Maybe that's a hint

> Most likely we get a request during that downtime

What request do you mean?

as-flow:
> We get 502 when proxy couldn't make a request to the instance. In most cases this means that nothing was serving on port 8080. Import restarts `writer` process, so we have some downtime. Most likely we get a request during that downtime

Forgot to ping you @UladzislauK-Writer

UladzislauK-Writer:
> Also I can't reproduce when running framework locally. Maybe that's a hint

Makes sense for local framework. Because we don't have a proxy layer that actually returns 502

> What request do you mean?

The one on the screenshot from the ticket, `/api/import`. 502 means API is down and Daytona can't reach it. Are we restarting the API process too? I would expect only AppRunner to be restarted. I mean, we definitely get the request since it is executed correctly. Might it be the case that we close the connection before we send a response and that gets interpreted as 502?

PR #1190, opened 2025-11-18
Vlad/ab 683

UladzislauK-Writer:
## Summary by CodeRabbit

* **New Features**
  * Execution journaling: records run metadata, trigger info, per-block outputs, and final status; runs alongside execution without changing user-visible behavior.
  * Journaling persists when the feature flag is enabled and storage is accessible.

* **Chores**
  * Added a read-only storage accessibility check before persistence.
  * Added a dedicated journal logger entry for diagnostics.

### DIFF
--- a/src/writer/journal.py
+++ b/src/writer/journal.py
@@ -39,7 +39,7 @@ class JournalRecord:
             self.trigger["component"]["id"] = graph.get_start_nodes()[0].id
 
         if "API" in title:
-            if component.type == "blueprints_crontrigger":
+            if getattr(component, "type", "") == "blueprints_crontrigger":
                 self.trigger["type"] = "Cron"
             else:
                 self.trigger["type"] = "API"
--- a/src/writer/journal.py
+++ b/src/writer/journal.py
@@ -1,0 +1,72 @@
+import logging
+from datetime import datetime, timezone
+from typing import TYPE_CHECKING, Any, Dict, Literal, Union
+
+from writer.core import Config
+from writer.keyvalue_storage import writer_kv_storage
+
+if TYPE_CHECKING:
+    from writer.blueprints import Graph, GraphNode
+    from writer.core_ui import Component
+
+
+logger = logging.getLogger("journal")
+
+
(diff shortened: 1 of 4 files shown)


src/writer/journal.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if "API" in title:
+            if getattr(component, "type", "") == "blueprints_crontrigger":
+                self.trigger["type"] = "Cron"
+            else:
+                self.trigger["type"] = "API"
+        elif "UI" in title:
+            self.trigger["type"] = "UI"
+        else:
+            self.trigger["type"] = "On demand"
```
UladzislauK-Writer, line 49: I don't like that I have to rely on the title here but anything else would require a significant refactor of our blueprint execution code

UladzislauK-Writer: For the context, title is that:
image
image
image

src/writer/journal.py
```diff
@@ -0,0 +1,72 @@
...
+            block_outputs[graph_node.id] = {"result": graph_node.result, "outcome": graph_node.outcome}
+        return {
+            "timestamp": self.started_at.isoformat(),
+            "instanceType": self.instance_type,
+            "trigger": self.trigger,
+            "blockOutputs": block_outputs,
+        }
+
+    def construct_key(self) -> str:
+        return f"wf-journal-{self.instance_type[0]}-{int(self.started_at.timestamp() * 1000)}"
```
bybash, line 65: Do we need that 1000?
My concern is making the key long.

bybash (commented):
Approved with one suggestion.

src/writer/journal.py
UladzislauK-Writer: We don't have any logic to act on key collisions so I decided to be on the safer side. This is ms, we probably can use seconds instead

bybash: I think that's a very small chance to have key collision and to be fair, it is an acceptable trade-off.

bybash (approved):

PR #1191, opened 2025-11-18
feat: expose API for KV access

UladzislauK-Writer:


src/writer/serve.py
```diff
@@ -290,6 +293,38 @@ async def autogen(requestBody: AutogenRequestBody, request: Request):
...
+    @app.post("/api/data/retrieve")
+    async def retrieve_data(requestBody: RetrieveDataRequestBody) -> RetrieveDataResponseBody:
+        from writer.keyvalue_storage import writer_kv_storage
+
+        all_keys = writer_kv_storage.get_data_keys()
+
+        keys_to_fetch = []
+        for key in all_keys:
+            if key in requestBody.skip_keys:
+                continue
```
UladzislauK-Writer, line 305: This will allow us to supply keys from FE that are already loaded and not fetch them again

src/writer/serve.py
```diff
@@ -290,6 +293,38 @@ async def autogen(requestBody: AutogenRequestBody, request: Request):
             agent_token_header
             )
 
+    @app.post("/api/data/retrieve")
+    async def retrieve_data(requestBody: RetrieveDataRequestBody) -> RetrieveDataResponseBody:
+        from writer.keyvalue_storage import writer_kv_storage
```
UladzislauK-Writer, line 298: I made these endpoints a bit more complex and not just 1-to-1 API wrappers because I didn't want to send a lot of requests through all the proxy layers

bybash (approved):

PR #1199, opened 2025-11-20
feat(UI): Journal Triggers [AB-719]

bybash:
https://writerai.atlassian.net/browse/AB-719
https://www.loom.com/share/5ed7806e06f04bfe96b8ad84388b47b6

### DIFF
--- a/src/ui/src/builder/journal/BuilderJournalSearchbar.vue
+++ b/src/ui/src/builder/journal/BuilderJournalSearchbar.vue
@@ -2,28 +2,38 @@
 	<div class="BuilderJournalSearchbar">
 		<WdsTextInput v-model="searchText" left-icon="search" />
 		<div class="BuilderJournalSearchbar__dropdown">
-			<WdsDropdownMenu
-				:selected="statuses"
+			<WdsSelect
+				v-model="statuses"
 				:options="statusOptions"
 				:enable-multi-selection="true"
-				@select="onStatusesChange"
-			></WdsDropdownMenu>
+				placeholder="Status"
+			/>
 		</div>
 		<div class="BuilderJournalSearchbar__dropdown">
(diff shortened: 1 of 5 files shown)


src/writer/blueprints.py
```diff
@@ -795,6 +795,8 @@ def _execute(self, executor: ThreadPoolExecutor, abort_event: threading.Event) -
                         entry_type="info",
                         exit="return"
                     )
+                    
+                    journal_record.save(result="success")
```
bybash, line 799: @UladzislauK-Writer this one I tried with simple trigger and return value in BP.
And it wasn't put into KV logs because that path using this return and it's not hitting the save to journal line on 808.

src/ui/src/builder/journal/BuilderJournalSearchbar.vue
```diff
@@ -40,25 +50,39 @@ const triggerOptions = [
...
-	statuses.value = val;
-}
-function onTriggersChange(val: string[]) {
-	triggers.value = val;
+
+const hasActiveFilters = computed(() => {
+	return (
+		(searchText.value && searchText.value.length > 0) ||
+		(statuses.value && statuses.value.length > 0) ||
+		(triggers.value && triggers.value.length > 0)
```
madeindjs, line 58:
```suggestion
		searchText.value ||
		statuses.value?.length ||
		triggers.value?.length
```

madeindjs (approved):

UladzislauK-Writer (approved):

src/ui/src/builder/journal/BuilderJournalSearchbar.vue
bybash: Yeap, I'll change it on next PR 👍

PR #1201, opened 2025-11-20
feat(UI): instance type filter [AB-721]

bybash:
https://writerai.atlassian.net/browse/AB-721
https://www.loom.com/share/5b1ab0d333f246eb9b28ab686a723f21

## Summary by CodeRabbit

* **New Features**
  * Added instance type filtering to journal entries, allowing filtering by Editor or Agent types.
  * Instance type labels now displayed with each journal entry for better visibility.

* **Style**
  * Improved code quality with stricter equality checks for keyboard event handling.

### DIFF
--- a/src/ui/src/builder/BuilderJournal.vue
+++ b/src/ui/src/builder/BuilderJournal.vue
@@ -1,9 +1,7 @@
 <template>
 	<div class="BuilderJournal">
 		<BuilderJournalHeader
-			v-model:search="searchText"
-			v-model:statuses="selectedStatuses"
-			v-model:triggers="selectedTriggers"
+			v-model:filters="filters"
 			@refresh="loadEntries"
 			@clear="deleteEntries"
 			@download="downloadAsJson"
@@ -26,6 +24,7 @@ import { convertAbsolutePathtoFullURL } from "@/utils/url";
 import { useToasts } from "./useToast";
 import { Component, WriterComponentDefinition } from "@/writerTypes";
 import injectionKeys from "@/injectionKeys";
+import type { JournalFilters } from "./journal/journalTypes";
 
 defineOptions({
 	name: "BuilderJournal",
@@ -59,15 +58,19 @@ export type RawJournalEntry = {
 
 export type JournalEntry = RawJournalEntry & {
 	title: string;
+	instanceTypeLabel: string;
 	component: Component;
 	componentDefinition: WriterComponentDefinition;
 };
 
 const rawEntries = ref<Record<string, RawJournalEntry>>({});
 
-const searchText = ref("");
-const selectedStatuses = ref<string[]>([]);
-const selectedTriggers = ref<string[]>([]);
+const filters = ref<JournalFilters>({
+	search: "",
+	statuses: [],
+	triggers: [],
+	instanceTypes: [],
+});
 
 const entries = computed<Record<string, JournalEntry>>(() => {
 	return Object.fromEntries(
@@ -82,25 +85,43 @@ const entries = computed<Record<string, JournalEntry>>(() => {
 					? component.content.key
 					: component.content.alias || componentDefinition.name;
 
-			return [key, { ...entry, title, component, componentDefinition }];
+			const instanceTypeLabel =
+				entry.instanceType.charAt(0).toUpperCase() +
+				entry.instanceType.slice(1);
+
+			return [
+				key,
+				{
+					...entry,
+					title,
+					instanceTypeLabel,
+					component,
+					componentDefinition,
+				},
+			];
 		}),
 	);
 });
 
 const filteredEntries = computed<Record<string, JournalEntry>>(() => {
-	const searchTextLower = searchText.value.toLowerCase();
+	const searchTextLower = filters.value.search.toLowerCase();
 	return Object.fromEntries(
-		Object.entries(entries.value).filter(([_, entry]) => {
+		Object.entries(entries.value).filter(([_key, entry]) => {
 			const searchMatch =
-				searchText.value === "" ||
+				filters.value.search === "" ||
 				entry.title.toLowerCase().includes(searchTextLower);
 			const statusMatch =
-				selectedStatuses.value.length === 0 ||
-				selectedStatuses.value.includes(entry.result);
+				filters.value.statuses.length === 0 ||
+				filters.value.statuses.includes(entry.result);
 			const triggerMatch =
-				selectedTriggers.value.length === 0 ||
-				selectedTriggers.value.includes(entry.trigger.type);
-			return searchMatch && statusMatch && triggerMatch;
+				filters.value.triggers.length === 0 ||
+				filters.value.triggers.includes(entry.trigger.type);
+			const instanceTypeMatch =
+				filters.value.instanceTypes.length === 0 ||
+				filters.value.instanceTypes.includes(entry.instanceType);
+			return (
+				searchMatch && statusMatch && triggerMatch && instanceTypeMatch
+			);
 		}),
 	);
 });
--- a/src/ui/src/builder/journal/BuilderJournalEntry.vue
+++ b/src/ui/src/builder/journal/BuilderJournalEntry.vue
@@ -88,7 +88,7 @@ const { formattedDate, formattedTime } = useDateTimeFormatter(dateObj, {
 }
 
 .BuilderJournalEntry__summary__instanceType {
-	font-size: 0.75rem;
+	font-size: 12px;
 	color: var(--wdsColorGray5);
 }
 
--- a/src/ui/src/builder/journal/BuilderJournalEntry.vue
+++ b/src/ui/src/builder/journal/BuilderJournalEntry.vue
@@ -4,8 +4,13 @@
 			<BuilderJournalEntryResult
 				:result="journalEntry.result"
 			></BuilderJournalEntryResult>
-			<div class="BuilderJournalEntry__summary__component">
-				{{ journalEntry.title }}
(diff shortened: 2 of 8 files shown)


as-flow:
@coderabbitai review

src/ui/src/builder/journal/BuilderJournalSearchbar.vue
```diff
@@ -33,32 +41,29 @@
...
+	STATUS_OPTIONS,
+	TRIGGER_OPTIONS,
+	INSTANCE_TYPE_OPTIONS,
+} from "./journalConstants";
+import type { JournalFilters } from "./journalTypes";
 
-const searchText = defineModel("search", { type: String });
-const statuses = defineModel("statuses", { type: Array as PropType<string[]> });
-const triggers = defineModel("triggers", { type: Array as PropType<string[]> });
+const filters = defineModel<JournalFilters>("filters", { required: true });
```
madeindjs: It's better to rely on runtime check

```suggestion
const filters = defineModel("filters", { type: Object as PropType<JournalFilters>, required: true });
```

src/ui/src/builder/journal/journalConstants.ts
madeindjs, line 1: I feel that it's better to keep it in `src/ui/src/builder/journal/BuilderJournalSearchbar.vue` since it's only used there. You can define a non-setup `script` that you can import in the same way in `src/ui/src/builder/journal/journalTypes.ts`

Also, I prefer `Object.freeze` instead of `as const` because it does runtime check

```vue
<script lang="ts">
import type { WdsDropdownMenuOption } from "@/wds/WdsDropdownMenu.vue";

export const STATUS_OPTIONS = Object.freeze<WdsDropdownMenuOption[]>([
	{ value: "success", label: "Success" },
	{ value: "error", label: "Error" },
	{ value: "stopped", label: "Stopped" },
]);

export const TRIGGER_OPTIONS = Object.freeze<WdsDropdownMenuOption[]>([
	{ value: "On demand", label: "On demand" },
	{ value: "UI", label: "UI" },
	{ value: "API", label: "API" },
	{ value: "Cron", label: "Scheduled" },
]);

export const INSTANCE_TYPE_OPTIONS = Object.freeze<WdsDropdownMenuOption[]>([
	{ value: "editor", label: "Editor" },
	{ value: "agent", label: "Agent" },
]);
</script>
```

src/ui/src/builder/journal/journalTypes.ts
```diff
@@ -0,0 +1,17 @@
...
+// Type helpers
+export type StatusValue = (typeof STATUS_OPTIONS)[number]["value"];
+export type TriggerValue = (typeof TRIGGER_OPTIONS)[number]["value"];
+export type InstanceTypeValue = (typeof INSTANCE_TYPE_OPTIONS)[number]["value"];
+
+export interface JournalFilters {
+	search: string;
+	statuses: string[];
+	triggers: string[];
+	instanceTypes: string[];
```
madeindjs:
```suggestion
	statuses: StatusValue[];
	triggers: TriggerValue[];
	instanceTypes: InstanceTypeValue[];
```

src/ui/src/builder/BuilderJournal.vue
```diff
@@ -59,15 +58,19 @@ export type RawJournalEntry = {
...
 
-const searchText = ref("");
-const selectedStatuses = ref<string[]>([]);
-const selectedTriggers = ref<string[]>([]);
+const filters = ref<JournalFilters>({
+	search: "",
+	statuses: [],
+	triggers: [],
+	instanceTypes: [],
+});
```
madeindjs, line 73: It's more performant to have multiples `ref` than a big one.

src/ui/src/builder/journal/BuilderJournalSearchbar.vue
```diff
@@ -1,22 +1,30 @@
...
+				:options="TRIGGER_OPTIONS"
 				:enable-multi-selection="true"
 				placeholder="Trigger"
 			/>
 		</div>
+		<div class="BuilderJournalSearchbar__dropdown">
+			<WdsSelect
+				v-model="filters.instanceTypes"
+				:options="INSTANCE_TYPE_OPTIONS"
+				:enable-multi-selection="true"
```
madeindjs: It's equivalent to
```suggestion
				enable-multi-selection
```

src/ui/src/builder/journal/BuilderJournalHeader.vue
```diff
@@ -42,16 +38,14 @@
...
 
 const emit = defineEmits({
 	refresh: () => true,
 	clear: () => true,
 	download: () => true,
 });
-const searchText = defineModel("search", { type: String });
-const statuses = defineModel("statuses", { type: Array as PropType<string[]> });
-const triggers = defineModel("triggers", { type: Array as PropType<string[]> });
+const filters = defineModel<JournalFilters>("filters", { required: true });
```
madeindjs, line 48: Why do you prefer a large prop to multiple ones? It's worse in terms of performances.

src/ui/src/builder/journal/BuilderJournalEntry.vue
```diff
@@ -67,16 +72,26 @@ const { formattedDate, formattedTime } = useDateTimeFormatter(dateObj, {
...
+	gap: 0px;
 	justify-content: center;
+}
+
+.BuilderJournalEntry__summary__component {
 	font-weight: 500;
 }
 
+.BuilderJournalEntry__summary__instanceType {
+	font-size: 0.75rem;
```
madeindjs: We try to avoid `rem` and use `px` instead. I know we still have legacy `rem` in some places, but let's try to get rid of them.
```suggestion
	font-size: 12px;
```

src/ui/src/builder/journal/BuilderJournalHeader.vue
bybash: Good question, from my perspective why I do like object props and prefer it because:
- Data naturally belongs together as a domain object (filter)
- When passing data through multiple component layers (Journal -> Header -> Filter)
- Better code organization and readability

Also before committing, as a React person (single objects are better in latest version), I checked again if this is not a good practice in Vue. But my findings showed me that's not a performance breaker in Vue too.
Here is what I read:

> Vue 3's reactivity system (based on Proxies) tracks dependencies at the property level, not at the object level. Whether you pass { name, age, email } as one prop or as three separate props, Vue still tracks each property individually for changes. The reactivity overhead is essentially the same.

src/ui/src/builder/journal/BuilderJournalSearchbar.vue
bybash: In runtime, does it check the object's shape and props, or just type check if it is an object?

src/ui/src/builder/journal/journalConstants.ts
bybash: Great comment, you're right we can have them in one file but it is better to keep them in a separate file for better readability and maintainability. 
Freeze prevents runtime mutations, not does runtime checks. 
I don't think, that's something we need to worry about but inside the codebase I see we used freeze, so I'll follow the pattern 👍 
Will make that freeze changes in next PR.

src/ui/src/builder/BuilderJournal.vue
bybash: What I read about Vue3 is, it supports object prop level reactivity and there's no performance change. But if you direct me something I can read and see the point, I'd like to check that.

bybash (commented):
@madeindjs thanks for the reviews, could you take a look again? 🙏

src/ui/src/builder/journal/BuilderJournalSearchbar.vue
```diff
@@ -33,32 +41,29 @@
...
-	{ value: "error", label: "Error" },
-	{ value: "stopped", label: "Stopped" },
-];
-const triggerOptions = [
-	{ value: "On demand", label: "On demand" },
-	{ value: "UI", label: "UI" },
-	{ value: "API", label: "API" },
-	{ value: "Cron", label: "Scheduled" },
-];
+import { computed } from "vue";
```
bybash:
```suggestion
import { computed, PropType } from "vue";
```

src/ui/src/builder/BuilderJournal.vue
madeindjs: The Vue.js logic of composability is to split references into smaller pieces.

Using big objects states some flaws:

1. You need to unwrap things in the template
```diff
- <Component :search />
+ <Component :search="filters.value.search" />
```
2. You need to watch `deep: true` to watch changes by references
2. `computed`/`watch` will execute every time any key object change
```ts
computed(() => filter.value.statuses.filter(/*whatever*/))
filters.value.search = "foo"// invalidate computed cache
```
4. A ref on a nested object is more expensive than a scalar value.

It's micro-optimizations, and I think it's fine for your PR, but I think we have to avoid generalizing this pattern.

src/ui/src/builder/journal/BuilderJournalHeader.vue
madeindjs: The thing is, Vue.js tracks changes and re-renders only what is needed.

For instance, this might re-render if `filter.statuses` changes. 
```vue
<WdsTextInput v-model="filters.search" left-icon="search" />
```

`search` might change on every keystroke and invalidate everything, and force everything to re-render.


Using `const searchText = defineModel("search", { type: String })` is more efficient because it's a simple string that emits a scalar value that can trigger changes when text changes. I agree it's uglier to have many props, but easier to understand the data flow.

madeindjs (approved):

PR #1206, opened 2025-11-21
feat: separate search / filters and go to block [AB-749] [AB-750]

bybash:
https://www.loom.com/share/1372475cdf644cb6bd8bd846bc40e37b

## Summary by CodeRabbit

* **New Features**
  * Live search binding on the journal header for entry searching.
  * Block navigation from entry details (navigate to a block and close the drawer).

* **UI Improvements**
  * Centered loading indicator and text while entries load.
  * Interactive, keyboard-accessible block headers with hover affordances.
  * Header spacing/responsive tweaks; clearing filters now also clears search.

### DIFF
--- a/src/writer/journal.py
+++ b/src/writer/journal.py
@@ -43,6 +43,7 @@ class JournalRecord:
         else:
             self.trigger["component"]["type"] = "block"
             component = graph.get_start_nodes()[0].component
+            self.trigger["component"]["id"] = component.id
             self.trigger["component"]["title"] = self._get_block_name(component)
 
         if "API" in title:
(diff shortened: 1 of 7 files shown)


src/writer/journal.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         else:
             self.trigger["component"]["type"] = "block"
             component = graph.get_start_nodes()[0].component
+            self.trigger["component"]["id"] = component.id
```
bybash, line 46: @VladKrasny for the API calls and Cron, we need to add the id as well, otherwise FE is crashing while listing blocks.

UladzislauK-Writer: wrong Vlad tagged :smile: , but LGTM

UladzislauK-Writer (approved):

src/writer/journal.py
bybash: Ahuahuha my bad, thanks 🙏

PR #1207, opened 2025-11-21
feat(writer): introduce If-Else block - AB-605

madeindjs:
image

## Summary by CodeRabbit

* **New Features**
  * Added a compact single-line code editor variant for tighter, inline code display.
  * Introduced a new Eval field type for form/config inputs to support evaluable expressions.
  * Added an IfElse block to enable conditional logic and branching in workflows.

### DIFF
--- a/src/ui/src/builder/BuilderEmbeddedCodeEditor.vue
+++ b/src/ui/src/builder/BuilderEmbeddedCodeEditor.vue
@@ -5,6 +5,7 @@
 		:class="{
 			'BuilderEmbeddedCodeEditor--full': variant === 'full',
 			'BuilderEmbeddedCodeEditor--halfScreen': variant === 'half-screen',
+			'BuilderEmbeddedCodeEditor--singleLine': variant === 'single-line',
 		}"
 	>
 		<div ref="editorContainerEl" class="editorContainer"></div>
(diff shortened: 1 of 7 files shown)


src/ui/src/builder/BuilderEmbeddedCodeEditor.vue
```diff
@@ -28,10 +29,12 @@ const editorContainerEl = useTemplateRef("editorContainerEl");
 const resizeObserver = new ResizeObserver(updateDimensions);
 let editor: monaco.editor.IStandaloneCodeEditor = null;
 
+type Variant = "full" | "minimal" | "half-screen" | "single-line";
```
bybash:
```suggestion
type EditorVariant = "full" | "minimal" | "half-screen" | "single-line";
```

src/ui/src/builder/BuilderEmbeddedCodeEditor.vue
```diff
@@ -118,6 +141,19 @@ onUnmounted(() => {
 	min-height: 50vh;
 }
 
+.BuilderEmbeddedCodeEditor--singleLine {
+	min-height: 35px;
+	max-height: 35px;
+	height: 35px;
+	padding: 8.5px 12px 8.5px 12px;
```
bybash, line 151: Why do we have this 8.5px values all around the codebase :) 
I mean we should stick with 4x or 8x for a better UI composition.

bybash: Not something to fix in this PR 👍

bybash (approved):

src/writer/blocks/ifelse.py
```diff
@@ -0,0 +1,87 @@
...
+                    lambda entry: self.runner.session.session_state.add_log_entry(
+                        "info", "Captured stdout", entry
+                    )
+                ),
+                use_logging_redirect(
+                    lambda entry: self.runner.session.session_state.add_log_entry(
+                        "info", "Captured logs", entry
+                    )
+                ),
+            ):
```
UladzislauK-Writer: We don't need to capture `print`s and `logger` calls with `eval` since we don't allow statements anymore

src/ui/src/builder/BuilderEmbeddedCodeEditor.vue
madeindjs: Yes, I think that something we just copy / pasted everywhere 🤔

PR #1225, opened 2025-11-27
feat: AB-755 | Add shared blueprint backend foundation

as-flow:
- Add block_library_db.py for in-memory snippet storage
- Add shared_blueprint.py blueprint class
- Add shared_blueprint_registry.py for registry and helpers
- Update blocks/__init__.py to register SharedBlueprint
- Update blueprints.py with run_blueprint_from_components and SharedBlueprint handling
- Add test_shared_blueprint.py tests

## Summary by CodeRabbit

* **New Features**
  * Added support for shared blueprints — reusable blueprint components that can be executed as nodes within workflows.
  * Added a shared blueprint management service to publish, list, retrieve, and version shared blueprints, with an in-memory implementation for testing and a swap-in mechanism for alternate implementations.


src/writer/blocks/sharedblueprint.py
```diff
@@ -0,0 +1,98 @@
...
+    def register(cls, type: str):
+        """Register the shared_blueprint type with its abstract template."""
+        super(SharedBlueprint, cls).register(type)
+        register_abstract_template(
+            type,
+            AbstractTemplate(
+                baseType="blueprints_node",
+                writer={
+                    "name": "Shared Blueprint",
+                    "description": "Executes a shared blueprint from the component tree.",
```
madeindjs, line 36: You might want a feature flag.

src/writer/blocks/__init__.py
```diff
@@ -63,3 +64,4 @@
 WriterWebSearch.register("blueprints_writerwebsearch")
 WriterKeyValueStorage.register("blueprints_writerkeyvaluestorage")
 WriterChatReplyWithToolConfig.register("blueprints_writerchatreplywithtoolconfig")
+SharedBlueprint.register("shared_blueprint")
```
madeindjs: We have a lot of frontend checks that evaluate `nodeType.startsWith('blueprints_')`. So, I'll suggest

```suggestion
BlueprintShared.register("blueprints_shared")
```

src/writer/blocks/sharedblueprint.py
```diff
@@ -0,0 +1,98 @@
...
+                    "category": "Logic",
+                    "toolkit": "blueprints",
+                    "fields": {
+                        "payload": {
+                            "name": "Payload",
+                            "desc": "The value specified will be available using the template syntax, e.g. @{payload}.",
+                            "default": "{}",
+                            "type": "Text",
```
madeindjs, line 45: I think you want `Object`, no?

```suggestion
                            "type": "Object",
```

src/writer/blocks/sharedblueprint.py
```diff
@@ -0,0 +1,98 @@
...
+        super().__init__(component, runner, execution_environment)
+        self.source_blueprint_id = component.content.get("sourceBlueprintId", "")
+
+    def run(self):
+        """Execute the source blueprint's components."""
+        if not self.source_blueprint_id:
+            raise ValueError("No source blueprint ID specified")
+
+        try:
+            payload = self._get_field("payload")
```
madeindjs, line 80: If you use object payload

```suggestion
            payload = self._get_field("payload", True)
```

src/writer/shared_blueprint_service.py
```diff
@@ -0,0 +1,408 @@
...
+    def publish(
+        self,
+        organization_id: str,
+        user_id: int,
+        request: PublishSharedBlueprintRequest,
+    ) -> PublishSharedBlueprintResponse:
+        """
+        Publish a shared blueprint (create new or add version to existing).
+        """
+        pass
```
madeindjs, line 164: It is not possible to publish an application from the public endpoint ([doc](https://dev.writer.com/api-reference/application-api/)). I think it'll need to be done from the frontend.

src/writer/blocks/sharedblueprint.py
as-flow: @madeindjs I didn't add feature flags for the backend code because I want to maintain backwards compatibility e.g. if someone enabled the feature flag and created a shared blueprint but then disabled the flag - I don't want their app to get into a bad state.

src/writer/blocks/sharedblueprint.py
as-flow: @madeindjs No, because the payload can either be a regular string, or a json payload inside a string `"{}"`. The same pattern is used for the Run Blueprint block

src/writer/shared_blueprint_service.py
as-flow: Makes sense. Let's tackle this as part of a separate ticket since this is just the mocked implementation.

src/writer/blocks/sharedblueprint.py
madeindjs: We can't display a non-functional block to the customer. I think we must have feature flags.

If the feature flag is removed for whatever reason, we display a fallback node that indicates that the block wasn't found, so it becomes easy for the user to spot that he missed the flag.

That's what we always do for every beta block.

as-flow: Okay I'll update for consistency then

PR #1234, opened 2025-12-02
feat: AB-755 | Add shared blueprints core functionality

as-flow:
- Add SharedBlueprint block class for executing shared blueprints
- Add drag/drop handling for shared blueprints in canvas
- Add shared blueprints section in sidebar component tree
- Add shared blueprints category in toolkit
- Update component description for shared blueprint names

## Summary by CodeRabbit

* **New Features**
  * Shared blueprints UI: sidebar sections with per-section add buttons, per-item branches, and empty-state messaging
  * Create, activate and delete shared blueprints with confirmation and success/error toasts
  * Shared-blueprint execution support so workflows can run blueprint payloads

* **Improvements**
  * Drag-and-drop can initialize new items from a shared blueprint source
  * Toolbar shows shared-blueprint notice and restricts certain blocks when editing shared blueprints
  * Component labels and tool lists surface source-blueprint names; creation is tracked with a new event
  * Content initialization now preserves existing values and only fills missing defaults
* **Bug Fixes**
  * Search/filter checks guard non-string content fields to avoid runtime errors


src/ui/src/composables/useWriterTracking.ts
```diff
@@ -38,7 +38,8 @@ type WriterTrackingEventName =
 	| "blueprints_run_succeeded"
 	| "blueprints_logs_trace_opened"
 	| "blueprints_new_added"
-	| "blueprints_block_output_copied";
+	| "blueprints_block_output_copied"
+	| "shared_blueprint_added";
```
madeindjs: I think you should prefix by `blueprints_`

```suggestion
	| "blueprints_shared_blueprint_added";
```

src/ui/src/core/templateMap.ts
```diff
@@ -151,7 +152,7 @@ const templateMap: TemplateMap = {
 	note: CoreNote,
 };
 
-const abstractTemplateMap: Record<string, AbstractTemplate> = {};
+const abstractTemplateMap = reactive<Record<string, AbstractTemplate>>({});
```
madeindjs, line 155: Why is this needed?

src/writer/blocks/__init__.py
```diff
@@ -63,3 +64,4 @@
 WriterWebSearch.register("blueprints_writerwebsearch")
 WriterKeyValueStorage.register("blueprints_writerkeyvaluestorage")
 WriterChatReplyWithToolConfig.register("blueprints_writerchatreplywithtoolconfig")
+SharedBlueprint.register("shared_blueprint")
```
madeindjs:
```suggestion
SharedBlueprint.register("blueprints_shared")
```

src/writer/blocks/sharedblueprint.py
```diff
@@ -0,0 +1,99 @@
...
+            source_blueprint = self.runner.session.session_component_tree.get_component(
+                self.source_blueprint_id
+            )
+            blueprint_name = source_blueprint.content.get("key", "Shared Blueprint") if source_blueprint else "Shared Blueprint"
+
+            return_value = self.runner.run_blueprint(
+                self.source_blueprint_id,
+                expanded_execution_environment,
+                title=f"Shared blueprint execution ({blueprint_name})"
+            )
```
madeindjs: I would omit the `__init__`

```suggestion      
    def run(self):
        """Execute the source blueprint's components."""
       source_blueprint_id = component.content.get("sourceBlueprintId", "")
        
        if not source_blueprint_id:
            raise ValueError("No source blueprint ID specified")

        try:
            payload = self._get_field("payload")
            expanded_execution_environment = self.execution_environment | {"payload": payload}

            # Get blueprint name for title (matching run_blueprint_by_key pattern)
            source_blueprint = self.runner.session.session_component_tree.get_component(
               source_blueprint_id
            )
            blueprint_name = source_blueprint.content.get("key", "Shared Blueprint") if source_blueprint else "Shared Blueprint"

            return_value = self.runner.run_blueprint(
                source_blueprint_id,
                expanded_execution_environment,
                title=f"Shared blueprint execution ({blueprint_name})"
            )
```

src/writer/blocks/__init__.py
as-flow: Whoops, good catch

src/ui/src/core/templateMap.ts
as-flow: Basically because shared blueprints are registered dynamically at runtime (not at import time like regular components), and the builder sidebar's categories computed property depends on getSupportedComponentTypes() which reads from this map.

Without reactive(), Vue wouldn't track mutations to the object, so adding a new shared blueprint wouldn't invalidate the computed property and the sidebar would never update to show the new component.

src/ui/src/builder/sidebar/BuilderSidebarToolkit.vue
```diff
@@ -77,11 +118,45 @@ function showAutogen() {
...
+	});
+	return allBlueprints.filter((c) => c.content?.isSharedBlueprint === true);
+});
+
+// Block types that should not be available when editing a shared blueprint
+const RESTRICTED_BLOCKS_IN_SHARED_BLUEPRINT = new Set([
+	"blueprints_runblueprint",
+	"blueprints_apitrigger",
+	"blueprints_uieventtrigger",
+	"blueprints_crontrigger",
```
madeindjs, line 148: Let's forbid shared blueprints relying on other shared blueprints, no ?

```suggestion
	"blueprints_crontrigger",
	"blueprints_shared",
```

madeindjs (approved):

src/ui/src/builder/sidebar/BuilderSidebarToolkit.vue
as-flow:
```suggestion
		// Don't show shared blueprints when editing a shared blueprint (no nesting for now)
		if (
			!isSharedBlueprintsEnabled.value ||
			isEditingSharedBlueprint.value
		) {
			return [];
		}
```

We already do that here by hiding the shared blueprints category when editing a shared blueprint

PR #1250, opened 2025-12-22
feat(writer): allow CRON trigger block to be run - WF-881

madeindjs:
## Summary by CodeRabbit

* **New Features**
  * Blueprints can be triggered via Cron scheduling in addition to API triggers.
  * Exposed Cron availability and trigger lookup for blueprints; payloads no longer require a trigger type.

* **Refactor**
  * Trigger detection unified to recognize both API and Cron triggers.
  * Run flow simplified to a single consistent execution path that auto-selects API (preferred) or Cron.

* **Bug Fixes**
  * Clearer error messages when no supported trigger is available.

### DIFF
--- a/src/writer/blueprints.py
+++ b/src/writer/blueprints.py
@@ -232,7 +232,6 @@ class BlueprintRunner:
     def run_blueprint_via_api(
         self,
         blueprint_id: str,
-        trigger_type: Literal["API", "Cron"],
         branch_id: Optional[str] = None,
         execution_environment: Optional[Dict[str, Any]] = None
     ):
@@ -240,7 +239,6 @@ class BlueprintRunner:
         Executes a blueprint by its key via the API.
 
         :param blueprint_id: The blueprint identifier.
-        :param trigger_type: The type of trigger ("API" or "Cron").
         :param branch_id: Optional branch ID to start execution from.
         :param execution_environment: The execution environment for
         the blueprint.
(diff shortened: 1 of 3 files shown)


src/writer/blueprints.py
```diff
@@ -203,7 +251,11 @@ def run_blueprint_via_api(
 
         trigger_id = branch_id
         if trigger_id is None:
-            trigger_id = self.get_blueprint_api_trigger(blueprint_id)
+            if trigger_type == "Cron" or not self.is_blueprint_api_available(blueprint_id):
+                trigger_id = self.get_blueprint_cron_trigger(blueprint_id)
+                trigger_type = "Cron"
+            else:
+                trigger_id = self.get_blueprint_api_trigger(blueprint_id)
```
UladzislauK-Writer:
```
        if trigger_id is None:
            if self.is_blueprint_api_available(blueprint_id):
                trigger_type = "API"
            else:
                trigger_type = "Cron"
        else:
            trigger_type = "Cron"
```

You can do

```
        if trigger_id is None:
            if self.is_blueprint_api_available(blueprint_id):
                trigger_type = "API"
            elif self.is_blueprint_cron_available(blueprint_id):
                trigger_type = "Cron"
            else:
                raise SomeError()
        else:
            trigger_type = "Cron"
```
            
  too if you want

UladzislauK-Writer: And get the actual branch id if you need

UladzislauK-Writer: It's mostly a matter of removing the `if` in the `serve.py` and the `trigger_type` param

madeindjs: Thank you for the suggestion; I fixed it in https://github.com/writer/writer-framework/pull/1250/changes/2ca3ee7d2d683ca8a58640e62b6d001aa40920d6 . It's much better now

UladzislauK-Writer (approved):

PR #1260, opened 2026-01-02
fix(BlueprintLibrary): fickering issue when searching - AB-850

as-flow:
## Summary by CodeRabbit

* **Bug Fixes**
  * Optimized Blueprint Library Panel loading: the loading indicator now appears only on initial load rather than on each search refinement, improving the speed and smoothness of library searches.
  * Blueprint Library Panel search is cleared when closing the panel.

### DIFF
--- a/src/ui/src/builder/panels/BuilderBlueprintLibraryPanel.vue
+++ b/src/ui/src/builder/panels/BuilderBlueprintLibraryPanel.vue
@@ -132,6 +132,7 @@ const blueprints = ref<
 	}>
 >([]);
 const isLoading = ref(false);
+const hasLoadedOnce = ref(false);
 
 const globalBlueprints = computed(() =>
 	blueprints.value.filter((bp) => bp.isReadonly === true),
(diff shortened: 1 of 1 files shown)


src/ui/src/builder/panels/BuilderBlueprintLibraryPanel.vue
```diff
@@ -186,6 +191,8 @@ watch(searchQuery, debouncedLoadBlueprints);
 
 function handleClose() {
 	isOpen.value = false;
+	hasLoadedOnce.value = false;
```
bybash, line 194: Do you need to set it false here? I think you don't clear blueprints value, so it could stay as it is.

bybash (approved):

src/ui/src/builder/panels/BuilderBlueprintLibraryPanel.vue
as-flow: Clear the search value you mean? I do clear it so I think this can stay

as-flow: We need this because we want to show the skeleton loader when the user closes and then re-opens the blueprint library modal

PR #1266, opened 2026-01-15
Ab 938 perficient add citations for the knowledge graph

neelasha-writer:
Simple changes to Ask Graph Question block to give user the option to add inline citations

## Summary by CodeRabbit

* **New Features**
  * Added inline graph citations to graph question responses so users can see which specific graph sources informed an answer; enabled by default.

* **Chores**
  * Minor project file formatting cleanup (non-functional).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "writer"
-version = "1.16.1"
+version = "1.26.0"
 description = "An open-source, Python framework for building feature-rich apps that are fully integrated with the Writer platform."
 authors = ["Writer, Inc."]
 readme = "README.md"
@@ -51,10 +51,11 @@ requests = ">= 2.31.0, < 2.33"
 uvicorn = ">= 0.20.0, < 1"
 watchdog = ">= 3.0.0, < 4"
 websockets = ">= 12, < 16"
-writer-sdk = ">= 2.3.1, < 3"
+writer-sdk = ">= 2.3.2, < 3"
 python-multipart = ">=0.0.7, < 1"
 orjson = "^3.11.0, <4"
 launchdarkly-server-sdk = "^9.12.0"
+writerai = "^4.0.1"
 
 [tool.poetry.group.build]
 optional = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "writer"
-version = "1.26.0"
+version = "1.16.1"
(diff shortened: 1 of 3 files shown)


UladzislauK-Writer (approved):

UladzislauK-Writer (approved):

src/writer/blocks/writeraskkg.py
```diff
@@ -93,15 +102,21 @@ def run(self):
...
 
             response = client.graphs.question(
                 graph_ids=graph_ids,
                 question=question,
                 stream=use_streaming,
-                subqueries=subqueries
+                subqueries=subqueries,
+                query_config= {
+                    "inline_citations": graph_citations
+                }
```
UladzislauK-Writer, line 115: This argument was added in the `writer-sdk==2.3.2` but we have `>=2.3.1` in the `pyproject.toml`. Can you update it to `>=2.3.2`, please?

UladzislauK-Writer (changes requested):

src/writer/blocks/writeraskkg.py
madeindjs: I think toggling the "Use citations" needs to transform the response of the block and include response text and citations array. So maybe something like `{"answer": "foo", "citations": [...]}`. It would require a small change here

```py
elif graph_citations:
  state_value = { "answer":  response.answer , "citations": ...}
  self._set_state(state_element, answer_so_far)
```

madeindjs (changes requested):
The result is not usable by the customer

Image

I think the end goal is to get the plain text version, and an array of dict listing the citations.

You can try to replicate this typical use case: the user displays a PDF that has been added to a knowledge graph, and the use of citations highlights the parts of the PDF that have been used for the response.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "writer"
-version = "1.16.1"
+version = "1.26.0"
```
madeindjs:
```suggestion
version = "1.16.1"
```

pyproject.toml
```diff
@@ -51,10 +51,11 @@ requests = ">= 2.31.0, < 2.33"
 uvicorn = ">= 0.20.0, < 1"
 watchdog = ">= 3.0.0, < 4"
 websockets = ">= 12, < 16"
-writer-sdk = ">= 2.3.1, < 3"
+writer-sdk = ">= 2.3.2, < 3"
 python-multipart = ">=0.0.7, < 1"
 orjson = "^3.11.0, <4"
 launchdarkly-server-sdk = "^9.12.0"
+writerai = "^4.0.1"
```
madeindjs:
```suggestion
```

PR #1272, opened 2026-01-26
feat: AB-938 - add citations to ask graph block

neelasha-writer:
Changes to response format when citations are not enabled to make sure existing agents don't break

## Summary by CodeRabbit

* **Changes**
  * Graph citations are now disabled by default when processing questions.
  * Result finalization is centralized for both streaming and non-streaming flows, yielding more consistent outputs.

* **Bug Fixes**
  * Fixed state handling to prevent incomplete or duplicate results during streaming and non-streaming processing.

### DIFF
--- a/src/writer/blocks/writeraskkg.py
+++ b/src/writer/blocks/writeraskkg.py
@@ -115,45 +115,42 @@ class WriterAskGraphQuestion(WriterBlock):
                 }
             )
 
-            answer_so_far = ""
-            result_dict = {}
-            citations_so_far = []
-
-            if use_streaming:
-                for chunk in response:
-                    try:
-                        delta_answer = chunk.model_extra.get("answer", "")
-                        answer_so_far += delta_answer
-
-                        if graph_citations:
-                            delta_sources = chunk.model_extra.get("sources", "")
-                            citations_so_far.extend(delta_sources)
-                            result_dict["answer"] = answer_so_far
-                            result_dict["citations"] = citations_so_far
-                        
-                    except json.JSONDecodeError:
-                        logging.error(
-                            "Could not parse stream chunk from graph.question")
-
-            else:
-                answer_so_far = response.answer
-            
-                if graph_citations:
-                    result_dict["answer"] = answer_so_far
-                    result_dict["citations"] = response.sources or []
-
-            if graph_citations:
-                self._set_state(state_element, result_dict)
-                self.result = result_dict
-            else: 
-                self._set_state(state_element, answer_so_far)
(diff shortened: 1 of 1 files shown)


src/writer/blocks/writeraskkg.py
```diff
@@ -124,30 +124,34 @@ def run(self):
...
-            self.result = answer_so_far
+                    result_dict["answer"] = answer_so_far
+                    result_dict["citations"] = response.sources or []
+
+            if graph_citations:
+                self._set_state(state_element, result_dict)
+                self.result = result_dict
+            else: 
+                self._set_state(state_element, answer_so_far)
+                self.result = answer_so_far
```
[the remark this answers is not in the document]
neelasha-writer: @UladzislauK-Writer @madeindjs I am a bit confused, I thought the definition of state element was not dependent on useStreaming, and was just used to update the agent state. I believe before this feature implementation, state element was used to set state regardless of whether streaming was enabled. Any insights into this suggestion would be awesome

src/writer/blocks/writeraskkg.py
```diff
@@ -124,30 +124,34 @@ def run(self):
...
-                        result_dict["answer"] = answer_so_far
 
                         if graph_citations:
                             delta_sources = chunk.model_extra.get("sources", "")
                             citations_so_far.extend(delta_sources)
+                            result_dict["answer"] = answer_so_far
                             result_dict["citations"] = citations_so_far
                         
-                        self._set_state(state_element, result_dict)
-
```
UladzislauK-Writer, line 135: Removing this breaks the "Use streaming" functionality. The whole purpose of it is to update the state variable as we iterate over the `response` which is arriving over the network in multiple parts. With it the user can get more "AI-chatbot-like" experience when you are not waiting for the whole answer to appear at once but instead see it being generated

Neelashab: @UladzislauK-Writer I see, so when streaming is enabled, I should include the set state inside the for loop?

UladzislauK-Writer: Yes, exactly

src/writer/blocks/writeraskkg.py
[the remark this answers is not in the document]
UladzislauK-Writer: Yes, `state_element` was used regardless of `use_streaming`

This is talking about the fact that you can have `use_streaming=False` and `state_element=None`. In this case execution will reach these lines and calling `self._set_state(None, ...)` will (most likely) throw an error

@neelasha-writer

